PR #8, opened 2019-11-14
Added nose & flake8 libs to test libraries

docent-net:
Those libs are needed for running tests. Not sure about versions though - suggested are latest atm and works fine with the project.


adejanovski:
thanks for the contribution @docent-net 👍

PR #27, opened 2019-12-03
Fix differential backups re-uploading files in S3

### LINKED ISSUE #25: Differential mode uploading all files from host
I've successfully been able to backup to S3 and restore my node data using medusa 0.3.0 but am finding that regardless whether I use mode=full or mode=differential all files are being uploaded from the host
In a simple test case where I simply backup the system keyspaces and run a full backup  first and then a differential backup immediately afterwards I get the following output

[2019-12-02 17:08:31] INFO: Backup done
[2019-12-02 17:08:31] INFO: - Started: 2019-12-02 17:07:28
                        - Started extracting data: 2019-12-02 17:07:32
                        - Finished: 2019-12-02 17:08:31
[2019-12-02 17:08:31] INFO: - Real duration: 0:00:58.868730 (excludes time waiting for other nodes)
[2019-12-02 17:08:31] INFO: - 376 files, 283.62 KB
[2019-12-02 17:08:31] INFO: - 376 files copied from host
[2019-12-02 17:08:31] INFO: - 0 copied from previous backup (test2)

adejanovski:
Fixes #25 
Differential backups in S3 are currently re-uploading everything at each backup, because hash comparison is broken (2 ways) : it compares hashes with different encodings and we haven't implemented multipart hashes.

This PR now computes multipart hashes in case the file size is over the threshold, and it compares md5 hashes with both encodings to cover all seen cases so far with libcloud.


medusa/backup.py
```diff
@@ -125,8 +152,16 @@ def _make_manifest_object(self, path_prefix, cached_item):
...
 
     def files_are_different(self, src, cached_item):
+        md5_hash = generate_md5_hash(src)
+        multipart_md5_hash = md5_hash
+        if src.stat().st_size >= int(self._storage_config.multi_part_upload_threshold):
+            multipart_md5_hash = md5_multipart(src, 8)
+
         return (src.stat().st_size != cached_item['size']
-                or (generate_md5_hash(src) != cached_item['MD5'] and self._storage_provider != Provider.LOCAL))
+                or (md5_hash != cached_item['MD5']
```
rzvoncek: This looks like we could extract all of the hash comparing into a separate function called something like `hashhes_match`.  In that function we could then more comprehensively compare things.

medusa/backup.py
```diff
@@ -57,10 +57,35 @@ def generate_md5_hash(src, block_size=BLOCK_SIZE_BYTES):
...
+
+    multipart_hash = hashlib.md5(b''.join(hash_list)).hexdigest()
+
+    return '%s-%d' % (multipart_hash, len(hash_list))
+
+
+def md5_part(f, count):
+    hash_md5 = hashlib.md5()
+    eof = False
+    for i in range(count * 16):
```
rzvoncek: After chatting with Alex, I understand this function is meant to have a cap on how much it reads, so it should not greedily go towards the end of the file.
If this is really so, I suggest we drop the `count` as an argument, and hard-code it together with the 16 and 65536, and perhaps hide behind some "constant" variables.

medusa/backup.py
```diff
@@ -57,10 +57,35 @@ def generate_md5_hash(src, block_size=BLOCK_SIZE_BYTES):
     return base64_md5
 
 
+def md5_multipart(src, part_size):
```
rzvoncek: The `part_size` seems to be more of a `part_count`. But as I'm saying in other comments, I'd suggest dropping it.

adejanovski: it's actually a size, but I agree on dropping it to rely on a constant instead.

medusa/backup.py
adejanovski: done.

medusa/backup.py
adejanovski: I simplified the function, so extracting stuff into a different function would just bring most of it now.

Issue #31, opened 2019-12-09
Support S3 RGW (CephFS)


nicholasamorim:
The `.ini` file states that `storage_provider should be either of "local", "google_storage" or the s3_* values from`

However, I think the `s3_*` only applies to the classes that define AWS S3 regions and not `S3_RGW`.

nicholasamorim:
Ceph S3_RGW supported - merged in #32.

PR #32, opened 2019-12-09
Support for S3_RGW (CephFS S3-compatible API)

nicholasamorim:
Hi, this is a proposal to add S3_RGW to the list of supported providers.

I'm actually not sure how to add tests - or even if I need to so. I have been using in our setup and it's working fine.

Aims to solve #31

### DIFF
--- a/medusa/storage/s3_rgw.py
+++ b/medusa/storage/s3_rgw.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Spotify AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


nicholasamorim:
@adejanovski 

- fixed the hash comparison
- added `is_aws_s3`
- updated readme

medusa/storage/s3_rgw.py
```diff
@@ -0,0 +1,55 @@
...
+    def connect_storage(self):
+        with io.open(os.path.expanduser(self.config.key_file), 'r', encoding='utf-8') as json_fi:
+            credentials = json.load(json_fi)
+
+        driver = S3RGWStorageDriver(
+            host=self.config.host,
+            port=self.config.port,
+            region="default",
+            signature_version="4",
+            key=credentials['aws_access_key_id'],
```
adejanovski: It seems a bit confusing to use a credential file that mentions AWS although we're not using an AWS account (unless I'm missing something).
Is it a standard way of authenticating to CEPH using the S3 api? Are aws credentials involved in any way?

nicholasamorim: Good point, they strive to keep the nomenclature the same but I clearly forgot to remove the `aws_`. Done.

nicholasamorim:
@adejanovski done

Issue #29, opened 2019-12-09
3.4+ in setup.py is not true, at least 3.6 is required


nicholasamorim:
`setup.py` allows medusa to be installed in any python that is 3.4+. However, when trying to use medusa, a SyntaxError is raised on the f-string usage.

```
Traceback (most recent call last):
  File "/usr/local/bin/medusa", line 6, in <module>
    from medusa.medusacli import cli
  File "/usr/local/lib/python3.5/dist-packages/medusa/medusacli.py", line 33, in <module>
    import medusa.backup
  File "/usr/local/lib/python3.5/dist-packages/medusa/backup.py", line 33, in <module>
    from medusa.index import add_backup_start_to_index, add_backup_finish_to_index, set_latest_backup_in_index
  File "/usr/local/lib/python3.5/dist-packages/medusa/index.py", line 21, in <module>
    import medusa.storage
  File "/usr/local/lib/python3.5/dist-packages/medusa/storage/__init__.py", line 32, in <module>
    from medusa.storage.google_storage import GoogleStorage
  File "/usr/local/lib/python3.5/dist-packages/medusa/storage/google_storage.py", line 26, in <module>
    from medusa.storage.abstract_storage import AbstractStorage
  File "/usr/local/lib/python3.5/dist-packages/medusa/storage/abstract_storage.py", line 25, in <module>
    import medusa.storage.concurrent
  File "/usr/local/lib/python3.5/dist-packages/medusa/storage/concurrent.py", line 161
    return f"{size:.{decimal_places}f}{unit}"
```

The [blog post](https://thelastpickle.com/blog/2019/11/05/cassandra-medusa-backup-tool-is-open-source.html) at TLP does state:

> The command line tool that uses Python version 3.6 and needs to be installed on all the nodes you want to back up.

arodrime:
@adejanovski we need to decide if we should rather merge in https://github.com/thelastpickle/cassandra-medusa/pull/33 or rather fix this issue so python3.4.

@nicholasamorim Alex might be busy, but we will get back to this asap, thanks for the feedback.

arodrime:
I just saw the mention that we wrote on the blog post to use python3.6 min. With this publicised and the deb soon to be released, maybe python 3.6 is a good minimal version, maybe we don't have to support python 3.4. 

Still Ubuntu LTS doesn't offer python 3.6 in the default available packages I heard, which might annoy some people. Yet there are workarounds...

[...]

PR #33, opened 2019-12-09
Correctly set Python support to >=3.6

nicholasamorim:
Aims to solve #29

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,11 +35,11 @@ setuptools.setup(
         'Environment :: Console',
         'Intended Audience :: System Administrators',
         'License :: OSI Approved :: Apache Software License',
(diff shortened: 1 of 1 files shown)


arodrime:
@adejanovski was considering fixing medusa for Python3.4 instead.

Either way, I agree we should fix this mismatch, indeed. :(

adejanovski:
If we have a chance to make Medusa work with python 3.4 (and if indeed we just have a single line of code that's breaking it), we should fix it and keep our requirement to 3.4.
I'll try to take a stab at it and make the integration tests run on python 3.4.

adejanovski:
Closing as we'll try to support Python 3.4 instead (see #39).

Thanks

PR #36, opened 2019-12-10
Basic logging to file

nicholasamorim:
~If `--log-file` is passed, the log will _also_ go into a file.~

- Added logging to file.
- Added logging configuration in `medusa.ini`

### DIFF
--- a/medusa/medusacli.py
+++ b/medusa/medusacli.py
@@ -18,6 +18,7 @@ from gevent import monkey
 monkey.patch_all()
 import datetime
 import logging
+import logging.handlers
 import socket
 import click
 import sys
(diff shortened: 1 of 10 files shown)


adejanovski:
Looking at that PR now, thanks @nicholasamorim

medusa/medusacli.py
```diff
@@ -58,9 +58,14 @@ def configure_logging(verbosity, without_log_timestamp):
     else:
         log_format = '[%(asctime)s] %(levelname)s: %(message)s'
 
+    logging_handlers = [logging.StreamHandler()]
+    if log_file:
+        logging_handlers.append(logging.FileHandler(log_file))
```
adejanovski: Could we make this a `RotatingFileHandler` instead? It's easy to forget about the logs until they fill the disks ;)

adejanovski: http://www.blog.pythonlibrary.org/2014/02/11/python-how-to-create-rotating-logs/

medusa/medusacli.py
```diff
@@ -70,6 +75,7 @@ def configure_logging(verbosity, without_log_timestamp):
 
 @click.group()
 @click.option('-v', '--verbosity', help='Verbosity', default=0, count=True)
+@click.option('--log-file', help='Log to file as well')
```
adejanovski: Being able to enforce the logging it at runtime using an argument is a good thing to have, but we could maybe push this a little further by making it part of the configuration. What do you think?
We could have a `log_file=/path/to/logs` in the `medusa.ini` file that would allow to set logging to a file by default.

If you don't feel like doing this in this PR, it's fine, we can add it later as part of another PR.

nicholasamorim: I have added support to `logging` section in the configuration and removed the CLI argument to avoid ambiguity.

medusa/medusacli.py
nicholasamorim: Done!

medusa/medusacli.py
```diff
@@ -47,21 +48,41 @@
 pass_MedusaConfig = click.make_pass_decorator(medusa.config.MedusaConfig)
 
 
-def configure_logging(verbosity, without_log_timestamp):
+def configure_file_logging(config):
```
nicholasamorim: The reason I'm using 2 separate functions is because loading the config _already produces logging output_ so I couldn't find an elegant way of initializing both at the same time and not have the log losing the format.

nicholasamorim:
@adejanovski done.

adejanovski:
Cool contribution, thanks @nicholasamorim

Issue #34, opened 2019-12-09
Set nodetool parameters


bhagenbourger:
We need to set nodetool parameters like a password or a port. I suggest to add a nodetool section into the medusa.ini config file and use parameters defined into this section to build nodetool commands.
The nodetool section could have these parameters, only set parameters will be used including:
 - username
 - password
 - passwordFilePath
 - host
 - port

arodrime:
Sounds good to me. Yet we are using rather '_' than CamelCase. Also there is a section called `cassandra` with some cql_* configs, thus it would make sense to me to add up to this existing section:

To be clear, I think the [medusa-example.ini](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa-example.ini#L15-L20) could look like this:
```
[cassandra]
;stop_cmd = /etc/init.d/cassandra stop
;start_cmd = /etc/init.d/cassandra start
;config_file = <path to cassandra.yaml. Defaults to /etc/cassandra/cassandra.yaml>
;cql_username = <my_cql_username>
;cql_password = <my_cql_password>
;nodetool_username =  <my_nodetool_username>
;nodetool_ password =  <my_nodetool_ password>
;password_file_path = <path to nodetool password file>
;nodetool_host = <host name or IP to use for nodetool>
;nodetool_port = <port number to use for nodetool>
```

arodrime:
This was merged. Closing the issue.

Thanks @bhagenbourger, sorry about the delay (around Christmas it is a bit of a tricky period).

PR #37, opened 2019-12-11
Set nodetool parameters

bhagenbourger:
Aims to solve #34

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -87,6 +87,26 @@ class CqlSessionProvider(object):
             return CqlSession(session)
 
 
+class Nodetool(object):
+
+    def __init__(self, cassandra_config):
+        self._nodetool = ['nodetool']
+        if cassandra_config.nodetool_username is not None:
+            self._nodetool += ['-u', cassandra_config.nodetool_username]
+        if cassandra_config.nodetool_password is not None:
+            self._nodetool += ['-pw', cassandra_config.nodetool_password]
+        if cassandra_config.nodetool_password_file_path is not None:
+            self._nodetool += ['-pwf', cassandra_config.nodetool_password_file_path]
+        if cassandra_config.nodetool_host is not None:
+            self._nodetool += ['-h', cassandra_config.nodetool_host]
+        if cassandra_config.nodetool_port is not None:
+            self._nodetool += ['-p', cassandra_config.nodetool_port]
+
+    @property
+    def nodetool(self):
+        return self._nodetool
+
+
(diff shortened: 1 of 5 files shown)


medusa/cassandra_utils.py
```diff
@@ -483,7 +504,7 @@ def is_node_up(config, host):
     if int(config.cassandra.is_ccm) == 1:
         args = ['ccm', 'node1', 'nodetool']
     else:
-        args = ['nodetool', '-h', host]
+        args = Nodetool(config.cassandra).nodetool + ['-h', host]
```
arodrime: Is there a reason not to use the standard `self._nodetool.nodetool` here? (like for lines above) as we have:

```
self._nodetool = Nodetool(cassandra_config)
```

arodrime:
I don't have a cluster around but integration tests are passing. I'm ok to merge this once the small thing above is fixed.

```
1 feature passed, 0 failed, 0 skipped
14 scenarios passed, 0 failed, 0 skipped
251 steps passed, 0 failed, 0 skipped, 0 undefined
Took 7m53.828s
```

medusa/cassandra_utils.py
bhagenbourger: The method `is_node_up` is not member of `Cassandra` class. So I can't access to `self` here

arodrime: oh! I remember that. We never fixed it... It should be part of Cassandra, but then for now this is probably good enough.

bhagenbourger: ok, thanks for the review @arodrime

jfharden:
@arodrime Is this able to get merged? The issue was marked as resolved and we could really use this feature (we have the same problem with needing authentication credentials for nodetool)

arodrime:
@jfharden hello. I'm back from the winter break. Sorry about the delay there. I wanted to test it on a cluster and had some troubles running my test environment 🙃.

For what it's worth you can already install this version if you are in a hurry (or just willing to give this a try).

To install this version, you can run this on all nodes:

```
sudo apt-get install -y python3-pip && sudo pip3 install git+https://github.com/bhagenbourger/cassandra-medusa.git@set_nodetool_parameters --upgrade --force-reinstall 
```

jfharden:
@arodrime Thanks, I tried this branch and have it working with jmx authentication. Thanks @bhagenbourger

arodrime:
Let's merge this, if you tested it. It worked for me too (even though I did not have much time to push testing further). 

LGTM, merging.

PR #39, opened 2019-12-11
Fix compatibility with python3.5

bhagenbourger:
Aims to solve #29

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3.6
+    - name: Set up Python 3.5
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.5
     - name: Setup Java Action
       uses: actions/setup-java@v1 
       with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3.6
+    - name: Set up Python 3.5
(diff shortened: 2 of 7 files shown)


nicholasamorim:
I think `tox.ini` should be updated to support 3.4 so that the whole test suite can run on python 3.4

adejanovski:
agreed.
@bhagenbourger, could you change the base python version in `tox.ini` to be 3.4 instead of 3.6?
https://github.com/thelastpickle/cassandra-medusa/blob/master/tox.ini#L4

Thanks!

nicholasamorim:
@adejanovski now it says interpreter not found. I don't really know what/how GH is running these tests (never used GH workflows) but is 3.4 interpreter actually available?

bhagenbourger:
I modified `.github/workflows/ci.yml` to use python 3.4 and also the step name, but it's always the old step name which is displayed in action result logs. I try to fix to it today.

adejanovski:
I think we need to use pyenv in order to install other python versions that tox will be able to use. I'm working on this right now in another branch and will let you know if I manage to run tox with py34.
By default, only py27 and py36 will be installed on ubuntu Bionic, which is why tox can't find the other versions we're asking for.

bhagenbourger:
Following the documentation, only theses versions are available in github actions:
Python 2.7.16
Python 3.5.7
Python 3.6.9
Python 3.7.5
Python 3.8.0

Link to the documentation: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#ubuntu-1804-lts

adejanovski:
But testing with 3.5 might still not be enough for our target of supporting 3.4.
I think I'm about to have something that works and will run the tests on both 3.4 and 3.6.

bhagenbourger:
Since python 3.4 is not available in Github Actions and after a discussion with @adejanovski, minimum version supported is python3.5

adejanovski:
Thanks @bhagenbourger!
Merged

PR #51, opened 2020-01-09
fix flake issue - line is too long

arodrime:


arodrime:
This one is really trivial and was approved internally by Alex. Merging :).

PR #53, opened 2020-01-09
Use awscli for large files downloads

### LINKED ISSUE #52: Large files downloads eat up all memory for S3
As reported in #46, restoring backups with large files can end up failing because of memory usage.
It seems that the simple `blob.download()` method of libcloud tries to read the data in memory before writing it to disk.

As it was done for uploads, we need to rely on awscli for now in order to use multi part downloads on large files.

adejanovski:
Fixes #52

When using S3, files above the configured threshold (100MB by default) will be downloaded using awscli instead of libcloud.

### DIFF
--- a/medusa/storage/aws_s3_storage/awscli.py
+++ b/medusa/storage/aws_s3_storage/awscli.py
@@ -39,18 +39,27 @@ class AwsCli(object):
         self._env = dict(os.environ)
         return False
 
-    def cp(self, *, srcs, bucket_name, dest, max_retries=5):
+    def cp_upload(self, *, srcs, bucket_name, dest, max_retries=5):
         job_id = str(uuid.uuid4())
         awscli_output = "/tmp/awscli_{0}.output".format(job_id)
         objects = []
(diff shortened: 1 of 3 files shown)


adejanovski:
@jfharden, would you mind testing this branch to see if it fixes the issue with the large file downloads?

I successfully tested that awscli is used on files larger than 100MB, so it should do it.
If you throttled transfers in medusa.ini, the downloads will now be affected as well.

jfharden:
@adejanovski I've just started a restore running, it was around an hour last time before it hit a very large file so I'll update you later to let you know if this works for me or not.

jfharden:
It's looking good, I'm back on an m5.xlarge (so only 16GB of memory) and on this instance type it was failing after around 30GB dowloaded I recall, the current restore has around 400GB downloaded so far and the stderr log of the medusa job shows lots of files being retrieved as multi part.

Thanks so much for this!

jfharden:
Completed downloading all the files. Thanks!

PR #69, opened 2020-02-04
Fix sstableloader calls for clusters without authentication

adejanovski:
Fix schema recreation when restoring using the sstableloader
Unthrottle downloads for AWS S3


adejanovski:
@jfharden,

I spent some time on your issue with the sstableloader, and came out with a few fixes. You were right about the null password being a problem here.

Issue #90, opened 2020-02-28
medusa download fails due to missing parameter


chebelom:
`medusa download --backup-name download-broken --download-destination /mnt/temp/`

```
Traceback (most recent call last):
  File "/usr/local/bin/medusa", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/medusa/medusacli.py", line 152, in download
    medusa.download.download_cmd(medusaconfig, backup_name, Path(download_destination))
  File "/usr/local/lib/python3.5/dist-packages/medusa/download.py", line 75, in download_cmd
    download_data(config.storage, node_backup, download_destination)
TypeError: download_data() missing 1 required positional argument: 'destination'
```

arodrime:
Fixed as well. Closing

Issue #89, opened 2020-02-28
get_last_complete_cluster_backup fails if no complete backups are available


chebelom:
`medusa get-last-complete-cluster-backup`
```
[...]
Traceback (most recent call last):
  File "/usr/local/bin/medusa", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/medusa/medusacli.py", line 252, in get_last_complete_cluster_backup
    print(backup.name)
AttributeError: 'NoneType' object has no attribute 'name'

```

Issue #86, opened 2020-02-21
fqdn parameter from /etc/medusa/medusa.ini is not being read


thalesac:
it only works if I set it in the command like i.e.
```
medusa --fqdn hostname backup
```

thalesac:
(google storage)

thalesac:
this is probably due to this default -> https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/medusacli.py#L99

Since it get set to the default, it has precedence over the medusa.ini file.

arodrime:
This was fixed it seems. Closing

PR #91, opened 2020-02-28
fix multiple issues: #86,  #89, #90

chebelom:
Fix multiple issues:
- #86 `fqdn` not honored in config file
- #90 `medusa download` command does not work due to missing parameter
- #89 `medusa get_last_complete_cluster_backup` explodes if no complete cluster backup exists

### DIFF
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -18,6 +18,7 @@ import configparser
 import logging
 import os
 import pathlib
+import socket
 import sys
 
 import medusa.storage
@@ -80,6 +81,7 @@ def load_config(args, config_file):
         'multi_part_upload_threshold': 100 * 1024 * 1024,
         'secure': True,
         'aws_cli_path': 'aws',
+        'fqdn': socket.getfqdn(),
     }
 
     config['logging'] = {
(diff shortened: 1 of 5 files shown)


adejanovski:
Hi @chebelom,

thanks a lot for sending this PR.
Would you mind adding some tests (either unit or integration) to ensure that your fixes won't get butchered in the future?

The integration tests are built on top of [Behave](https://behave.readthedocs.io/en/latest/) and use the Gherkin syntax.
The download issue and the exploding get latest backup when there are none should be fairly easy to add to one of the existing test suites (you can pick the first one).

Let me know if you need some help.

chebelom:
Hi there,
I added a test to cover #90.

About #89:
in one of your integration tests you already check it.
https://github.com/thelastpickle/cassandra-medusa/blob/71d482b94f989b42e29213123dce552845c97a88/tests/integration/features/steps/integration_steps.py#L496
I just catch a `None` in the cli command. 

About  #86:
I do not know how to tackle this: if you think it requires a test, then I think I need some help :) .

adejanovski:
@chebelom,

LGTM. Could you please squash your commits into a single one before I merge the PR?
Thanks!

adejanovski:
Also, seems like there's a conflict now that I merged another PR which touched `config.py`.

adejanovski:
Thanks @chebelom! Merged 👍

PR #131, opened 2020-05-18
[WIP] Refactor checksum comparisons

### LINKED ISSUE #124: Refactor checksum comparisons to have them implemented in a single place.
There are three places at least where we deal with checksum comparisons to evaluate if files are the same.
We need to refactor this by having an implementation of the comparison per storage provider, and a single utility function to perform the check.

jsanda:
Fixes #124

There is still a lot of work to do but I wanted to open the PR and get some feedback. I have added a method named `files_are_different` in `abstract_storage.py`. It is intended to replace the `files_are_different` function in `backup.py`.  

I still need to update unit tests in `RestoreNodeTest` and remove unused code in `backup.py`. 

You will also see that the implementations of the `checksums_match` function currently call the helper functions in `backup.py`. If they are not used anywhere else, then it probably makes sense to move those functions closer to where they are used which will be in the storage providers.

I also need to refactor the `validate_manifest` method in `verify.py`. I have not started on it yet. It is a different use case that the `files_are_different` method and will most likely require another method in the storage drivers.

### DIFF
--- a/medusa/backup.py
+++ b/medusa/backup.py
@@ -136,11 +136,7 @@ class NodeBackupCache(object):
                 if self._storage_provider == Provider.GOOGLE_STORAGE or self._differential_mode is True:
                     cached_item = self._cached_objects.get(fqtn, {}).get(src.name)
 
-                if cached_item is None \
-                    or files_are_different(src,
-                                           cached_item,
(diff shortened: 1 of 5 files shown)


adejanovski:
@rzvoncek, this was done in your PR right? 
I guess we can merge this one then?

Issue #137, opened 2020-05-20
Implement gRPC service layer


jsanda:
There is ongoing discussion in #24 about using Medusa in Kubernetes. One of the things that needs to be done is to provide a _service_ layer, presumably HTTP-based. #88 was already opened for this and I started doing some investigation into creating a REST service for Medusa. I am instead going to build a gRPC service.

I think that RPC-style endpoints map more naturally to Medusa commands than a REST API would. This makes me think I will be able to develop a gRPC service layer more quickly than I could develop a REST service layer.

The goal of this ticket is to provide service endpoints that are be needed for Kubernetes use cases. 

The goal is **not** to simply expose all Medusa functionality, commands, etc via gRPC. Consider the `restore-cluster` command as an example. Medusa takes on the role of orchestrator with `restore-cluster`.  We do not need Medusa to assume the role of orchestrator in Kubernetes; therefore, the gRPC service does not need to provide an endpoint for `restore-cluster`.

The gRPC service also needs to expose a health check endpoint. The process for implementing a health check endpoint is described in https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-273) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-273

julienlau:
I don't know if auth design is layed out, but if so I'm wondering: will it be based on cassandra auth ? K8S auth ?
Will it also work without TLS ?

jsanda:
Closing since this is done.

PR #141, opened 2020-05-21
gRPC service layer

jsanda:
This PR is for #137.

There is currently only support for backups. I have tested it locally as well as in GCP. This initial commit also includes `Dockerfile.k8s` which is used to build an image for running the gRPC service in Kubernetes.

I created some K8S manifests for deploying a Cassandra cluster with Medusa in a sidecar container. I will put them in another repo and then link to them. I do not think it makes sense to include them here as they probably throw away work in the end.


jsanda:
> I created some K8S manifests for deploying a Cassandra cluster with Medusa in a sidecar container. I will put them in another repo and then link to them.

https://github.com/jsanda/cassandra-medusa-k8s

jsanda:
The primary motivation for implementing a gRPC service layer is for use in Kubernetes; however, this is nothing specific about k8s in this PR. The gRPC service can be used in any environment. In fact I used it locally for some early testing. The script for starting the gRPC server is `medusa/service/grpc_svc/server.py`.

The API for the service is defined by [medusa.proto](https://github.com/jsanda/cassandra-medusa/blob/grpc/medusa/service/grpc_svc/medusa.proto). It currently defines only two functions - `Backup` and `BackupStatus`. The purpose of this PR is not provide a gRPC API for all operations but rather only those needed in k8s. 

This PR introduces a new docker image, [k8s/Dockerfile](https://github.com/jsanda/cassandra-medusa/blob/grpc/k8s/Dockerfile) which is used in two scenarios:

* run gRPC server in sidecar container alongside Cassandra container in order to perform backups
* run an [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) to perform restores
    * `medusa/service/grpc_svc/restore.py` is the script that gets called to perform restores

This PR introduces a `grpc` config section that includes a flag to indicate it is enabled. When enabled,

* `sudo` is not used
* Calls to `nodetool` are replaced with calls to jolokia
* Calls to stop/start Cassandra are not made

These changes are needed for k8s. The usage of Jolokia is currently hard coded, but it needs to be parameterized. When Medusa is integrated with Cass Operator, calls to Jolokia will be replaced with calls to the management sidecar.

Testing so far has only been done with GCS.

jsanda:
This PR has been replaced by #223.

PR #145, opened 2020-05-26
Keep system keyspaces when user specifies ks/tables

### LINKED ISSUE #136: system_schema ks is not restored when restoring select ks/tables
When a list of keyspaces and/or tables is set in the restore-cluster command, Medusa will not restore the `system_schema` ks.
This ends up not creating the underlying schema of the sstables we're restoring.

arodrime:
Fixes #136.

Tested and working specifying a keyspace or a table.

### DIFF
--- a/medusa/restore_node.py
+++ b/medusa/restore_node.py
@@ -198,7 +198,7 @@ def invoke_sstableloader(config, download_dir, keep_auth, fqtns_to_restore, stor
 
 
 def keyspace_is_allowed_to_restore(keyspace, keep_auth, fqtns_to_restore):
-    if keyspace == 'system':
+    if keyspace == 'system' or keyspace == 'system_schema':
         return False
 
     if keyspace == 'system_auth' and keep_auth is True:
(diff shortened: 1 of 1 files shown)


arodrime:
When using sstable loader, we don't want to add those system keyspaces. And that's fine because we exclude the system keyspace in the `keyspace_is_allowed_to_restore()` function.

Yet I'm wondering if restoring other system keyspaces is ok with the sstable loader (ie `system_distributed` and `system_schema`. 

@adejanovski do you think restoring those 2 tables in the case we use a sstableloader is a problem, or not?

adejanovski-ds:
With the sstableloader, we should not restore the `system_schema` keyspace as the schema gets recreated using the backed up `schema.cql` file.

adejanovski:
@arodrime, with the sstableloader, we should not restore the system_schema keyspace as the schema gets recreated using the backed up schema.cql file.

arodrime:
After talking with Alex, I need to make a quick fix here still. Will do tomorrow.

arodrime:
@adejanovski
- tests are passing now ¯\_(ツ)_/¯
- manual testing make me confident as well

It should be easy enough to review quickly if you could. Then happy to merge on my side.

PR #156, opened 2020-06-05
Enforce tokens by modifying cassandra.yaml instead of setting env vars

### LINKED ISSUE #130: Apparent loss of data after backup/restore with medusa
We are currently evaluating cassandra-medusa to perform backup/restore of our cassandra clusters and it is working well so far.
In order to validate everything, we are performing some checks after restore and the result is quite disappointing.

The cinematic is the following :
• One existing cluster (name it A) with 3 DC that have data. Data is replicated on all DC.
• One new cluster with same topology but totally empty (name it B)

We launch backup on all nodes of 1 DC of A : everything is OK, our data are sent to S3. medusa verify --backup-name="backup" says it is ok.
here is the command we use :

```
medusa \
    -v \
    --fqdn='$(hostname -fA | cut -d" " -f2)' `# workarround to pass fqdn to medusa should be fixed in later version` \
    --ssh-username=cassandra \
    --ssh-key-file="ssh_key_name" \
 backup \
   --backup-name="backup_name" \
   --mode="full"

```

• We launch the restore on the new cluster, for a specific table
we use the following command :
```
medusa -v \
      --fqdn='$cassandra_hosts_fqdn' \
      --ssh-username=cassandra \
      --ssh-key-file="$ssh_key_name" \
  restore-cluster \
      --table='$target_table' \
      --backup-name="backup_name" \
      --temp-dir=/opt/apache-cassandra-3.11.2/data/tmp \
      --host-list /tmp/restore_mapping.txt \
      --verify \
      -y
```
• We perform a repair on this table (using reaper)
• Then we compare the number of lines between the table in cluster A and table in cluster B
Here is the result :

> Original table number of lines : 48 369 310 
> Restored table number of lines : 42 930 613

We are missing 6 millions lines....
Any idea what could cause this ? Did we misdone/forget something ?

Extra informations : 
TTL : no TTL set on data
Cluster : each cluster has 6 nodes each (same configuration terraformed)
vnodes : enabled

### LINKED ISSUE #133: Restore uses sstableloader with vnodes
Although we have the same number of nodes, when restore-cluster is invoked without a host list on a different cluster using vnodes, Medusa decides to use the sstableloader.
It shouldn't do so as we enforce the tokens and can push sstables directly, which is more efficient.

Relates to #130

### LINKED ISSUE #142: Properly enforce tokens on cross topology restores
Tokens require to be enforced when restoring a backup to a new cluster with different token allocation.
This is done by setting an environment variable and fails if the start command is different than `/etc/init.d/cassandra start`. This command though, will fail at putting the service into a proper state, which will create further problems.

We need to find a "clean" way of setting the tokens reliably for all scenarios.

adejanovski:
fixes #142
fixes #133 
fixes #130

### DIFF
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -233,7 +233,8 @@ class RestoreJob(object):
                 # ops might not be aware of the underlying behavior towards auth. Let's ask what to do...
                 really_keep_auth = None
                 while (really_keep_auth != 'Y' and really_keep_auth != 'n') and not self.bypass_checks:
-                    really_keep_auth = input('Do you want to skip restoring the system_auth keyspace and keep the credentials of the target cluster? (Y/n)')
+                    really_keep_auth = input('Do you want to skip restoring the system_auth keyspace and keep the'
+                                             + ' credentials of the target cluster? (Y/n)')
                 self.keep_auth = True if really_keep_auth == 'Y' else False
 
         if topology_matches:
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -51,15 +51,6 @@ def orchestrate(config, backup_name, seed_target, temp_dir, host_list, keep_auth
             logging.error(err_msg)
             raise Exception(err_msg)
 
-        if keep_auth:
-            really_keep_auth = "Y"
-            logging.info('system_auth keyspace will be left untouched on the target nodes')
-        else:
-            # ops might not be aware of the underlying behavior towards auth. Let's ask what to do...
-            really_keep_auth = None
-            while (really_keep_auth != 'Y' and really_keep_auth != 'n') and not bypass_checks:
-                really_keep_auth = input('Do you want to skip restoring the system_auth keyspace and keep the credentials of the target cluster? (Y/n)')
-
         storage = Storage(config=config.storage)
 
         try:
@@ -69,7 +60,7 @@ def orchestrate(config, backup_name, seed_target, temp_dir, host_list, keep_auth
             logging.error(err_msg)
             raise Exception(err_msg)
 
-        restore = RestoreJob(cluster_backup, config, temp_dir, host_list, seed_target, True if really_keep_auth == "Y" else False, verify,
+        restore = RestoreJob(cluster_backup, config, temp_dir, host_list, seed_target, keep_auth, verify,
                              pssh_pool_size, keyspaces, tables, bypass_checks, use_sstableloader)
         restore.execute()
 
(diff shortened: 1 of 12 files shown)


medusa/restore_cluster.py
```diff
@@ -34,17 +34,17 @@
...
         restore_start_time = datetime.datetime.now()
-        if seed_target is not None:
+        if seed_target is not None and host_list is None:
             keep_auth = False
 
         if seed_target is None and host_list is None:
-            err_msg = 'You must either provide a seed target or a list of host'
-            logging.error(err_msg)
-            raise Exception(err_msg)
+            retained_seed_target = socket.getfqdn()
```
arodrime, line 44: Why using retained_seed_target, if in that case seed_target is empty, why not just setting it directly instead of creating a new variable for the same thing?

medusa/restore_cluster.py
```diff
@@ -213,7 +214,7 @@ def _tokens_from_ringitem(ringitem):
 
         def _token_counts_per_host(tokenmap):
             for host, ringitem in tokenmap.items():
-                yield len(ringitem['tokens'])
+                return len(ringitem['tokens'])
```
arodrime, line 217: We don't actually need the `for` loop here after your update. We can just get the first host and return the length of its token list. I know the return breaks the loop, but it's probably a bit better to get the information directly.

medusa/restore_cluster.py
```diff
@@ -235,8 +236,8 @@ def _chunk(my_list, nb_chunks):
             target_tokens = {_tokens_from_ringitem(ringitem): host for host, ringitem in target_tokenmap.items()}
             backup_tokens = {_tokens_from_ringitem(ringitem): host for host, ringitem in tokenmap.items()}
 
-            target_tokens_per_host = set(_token_counts_per_host(tokenmap))
-            backup_tokens_per_host = set(_token_counts_per_host(target_tokenmap))
+            target_tokens_per_host = _token_counts_per_host(tokenmap)
+            backup_tokens_per_host = _token_counts_per_host(target_tokenmap)
```
arodrime, line 240: Ok, so what was done before was actually broken. Considering all nodes have the same number of tokens good enough for now.

medusa/restore_cluster.py
```diff
@@ -247,21 +248,32 @@ def _chunk(my_list, nb_chunks):
...
+                target_ringmap[token].append(host)
+
+            self.ringmap = backup_ringmap
+            i = 0
+            for token, hosts in backup_ringmap.items():
+                # take the node that has the same token list or pick the one with the same position in the map.
+                restore_host = target_ringmap.get(token, list(target_ringmap.values())[i])[0]
+                isSeed = True if socket.getfqdn(restore_host) in self._get_seeds_fqdn() else False
+                self.host_map[restore_host] = {'source': [hosts[0]], 'seed': isSeed}
+                i += 1
```
arodrime, line 276: This part above I don't really understand yet. I'll do testing / play around see if it works as expected.

medusa/restore_cluster.py
adejanovski: you're right, that part is fairly unclear...

The checks were done to identify when a remote restore was done to force the keep_auth value, but that's not relevant to do it this way as we can do remote restores without specifying a host list file.
I'll rework all this and ask the use if he wants to retain `system_auth` unless `--keep-auth` is set explicitly.

medusa/restore_cluster.py
arodrime: I could reproduce the bug in using a remote cluster and a `seed-host`. This code seems to fix the problem I was reproducing in master branch. This looks good to me.

PR #166, opened 2020-06-22
Fix gsutil upload/download 

house-of-vanity:
gsutil cp uploading/downloading issue. fix feeding plenty of files to gsutil's stdin which causes fail.

### DIFF
--- a/medusa/backup.py
+++ b/medusa/backup.py
@@ -31,7 +31,8 @@ from medusa.cassandra_utils import Cassandra
 from medusa.index import add_backup_start_to_index, add_backup_finish_to_index, set_latest_backup_in_index
 from medusa.monitoring import Monitoring
 from medusa.storage.s3_storage import is_aws_s3
-from medusa.storage import Storage, format_bytes_str, ManifestObject
+from medusa.storage.google_storage import GSUTIL_MAX_FILES_PER_CHUNK
+from medusa.storage import Storage, format_bytes_str, ManifestObject, divide_chunks
 
 
 class NodeBackupCache(object):
@@ -322,7 +323,11 @@ def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot
 
             manifest_objects = list()
             if len(needs_backup) > 0:
-                manifest_objects = storage.storage_driver.upload_blobs(needs_backup, dst_path)
+                # If there is a plenty of files to upload it should be
+                # splitted to batches due to 'gsutil cp' which
+                # can't handle too much source files via STDIN.
+                for src_batch in divide_chunks(needs_backup, GSUTIL_MAX_FILES_PER_CHUNK):
+                    manifest_objects += storage.storage_driver.upload_blobs(src_batch, dst_path)
 
             # Reintroducing already backed up objects in the manifest in differential
             for obj in already_backed_up:
--- a/medusa/download.py
+++ b/medusa/download.py
@@ -18,7 +18,8 @@ import json
 import pathlib
 import sys
 
-from medusa.storage import Storage
+from medusa.storage import Storage, divide_chunks
+from medusa.storage.google_storage import GSUTIL_MAX_FILES_PER_CHUNK
 
 
 def download_data(storageconfig, backup, fqtns_to_restore, destination):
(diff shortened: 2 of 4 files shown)


medusa/backup.py
```diff
@@ -385,8 +385,20 @@ def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot
 
         manifest_objects = list()
         if len(needs_backup) > 0:
-            manifest_objects = storage.storage_driver.upload_blobs(needs_backup, dst_path)
-
+            # If there is a plenty of files to upload it should be
```
adejanovski, line 388: Could we rework the code in order to use a function that will split the `needs_backup` list into chunks of a given size, and then loop through the chunks?
It would simplify the code and make it more readable.
Something like [this](https://www.geeksforgeeks.org/break-list-chunks-size-n-python/):

```
def divide_chunks(l, n): 
    for i in range(0, len(l), n):  
        yield l[i:i + n] 

...
MAX_FILES_PER_BACKUP_CHUNK = 64
...
backup_chunks = list(divide_chunks(needs_backup, MAX_FILES_PER_BACKUP_CHUNK))
files_uploaded = 0
for batch in backup_chunks:
  logging.debug("Uploading %s files to %s.", len(batch), dst_path)
  manifest_objects += storage.storage_driver.upload_blobs(batch, dst_path)
```

What do you think?

medusa/download.py
```diff
@@ -43,8 +43,21 @@ def download_data(storageconfig, backup, fqtns_to_restore, destination):
             subfolder.mkdir(parents=False)
 
         if len(srcs) > 0 and (len(fqtns_to_restore) == 0 or fqtn in fqtns_to_restore):
-            logging.info('Downloading backup data')
-            storage.storage_driver.download_blobs(srcs, dst)
+            logging.info('Downloading backup data - %s files.', len(srcs))
```
adejanovski, line 46: Same comment as for uploads.

medusa/backup.py
house-of-vanity: I got it. How do you think is it OK to move this function in medusa.storage and import when needed as auxiliary function?

house-of-vanity: Anyway I've done it somehow, please review and let me know.

medusa/storage/__init__.py
```diff
@@ -44,6 +44,11 @@
 INDEX_BLOB_WITH_TIMESTAMP_PATTERN = re.compile('.*(started|finished)_(.*)_([0-9]+).timestamp$')
 
 
+def divide_chunks(values, step=64):
```
adejanovski, line 47: Why set a default value for `step` here?
The method could be used for dealing with other things than splitting files to upload to GCP, where `64` would not make sense.
Since you already defined a constant for the GSUTIL chunk size and use it, I'd be keen to remove the default value to `step`.
Sounds good?

house-of-vanity: Ok, I've fixed this.

adejanovski:
@house-of-vanity, you still have a conflict in `backup.py` that needs fixing.
Also, could you squash all commits into a single one once you've rebased against `master`?

Thanks!

house-of-vanity:
> @house-of-vanity, you still have a conflict in `backup.py` that needs fixing.
> Also, could you squash all commits into a single one once you've rebased against `master`?
> 
> Thanks!

Ok, I'll do it soon . Git isn't my strong point, so it will take some time to understand how to do it.

house-of-vanity:
> @house-of-vanity, you still have a conflict in `backup.py` that needs fixing.
> Also, could you squash all commits into a single one once you've rebased against `master`?
> 
> Thanks!

It's done.

house-of-vanity:
> Did you get a chance to test this branch against an actual cluster and a GCS bucket?

Yes, I've tested it against my development cluster. Moreover reason of creating this PR is that i couldn't backup and restore my cluster due to issue.

dmitry-saprykin:
@adejanovski Hi, I have the same problem as @house-of-vanity with too many files for gsutil to process. This PR looks like merged and closed. However I cannot find any changes from it in the master branch.

Issue #164, opened 2020-06-19
Restoring issue. Medusa needs excessive file system permissions.


house-of-vanity:
Hi, I've faced an issue with restoring whole cluster.
When all the sstables are downloaded to host from bucket medusa tries to cleaning /srv/cassandra/commitlog by executing `sudo -u cassandra rm -rf /srv/cassandra/commitlog` which deletes entire directory not only the content. Due to permissions medusa fails this step. Note that cassandra doesn't create this directory themself and it's not assumed that cassandra has an access to this directory and only its content should be RW.
Maybe it's better to remove its content by executing `sudo -u cassandra rm -rf /srv/cassandra/commitlog/*` ?

```
[2020-06-19 14:03:59,662] INFO: Stopping Cassandra
[2020-06-19 14:03:59,674] DEBUG: Cleaning (/srv/cassandra/commitlog)
rm: cannot remove '/srv/cassandra/commitlog': Permission denied
Traceback (most recent call last):
  File "/usr/local/bin/medusa", line 33, in <module>
    sys.exit(load_entry_point('cassandra-medusa==0.6.0', 'console_scripts', 'medusa')())
  File "/usr/lib/python3/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/cassandra_medusa-0.6.0-py3.6.egg/medusa/medusacli.py", line 211, in restore_node
    verify, set(keyspaces), set(tables), use_sstableloader)
[...]

Issue #109, opened 2020-04-10
Problem with the restore 


TryToWork:
Hello, 

Restore process crashes due to the fact that we have a mount point.

```
[root@node1~]# df -h
Filesystem                                     Size  Used Avail Use% Mounted on
...
/dev/mapper/vg01-lv_cassandra_commitlog_tst     10G   48M   10G   1% /var/lib/cassandra/commitlog
/dev/mapper/vg01-lv_cassandra_savedcaches_tst  8.0G   33M  8.0G   1% /var/lib/cassandra/saved_caches
```

> rm: cannot remove ‘/var/lib/cassandra/commitlog’: Device or resource busy
Traceback (most recent call last):
  File "/opt/rh/rh-python36/root/usr/bin/medusa", line 11, in <module>
    sys.exit(cli())
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/medusa/medusacli.py", line 212, in restore_node
    verify, set(keyspaces), set(tables), use_sstableloader)
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/medusa/restore_node.py", line 48, in restore_node
    keyspaces, tables)
[...]

PR #168, opened 2020-06-22
Do not delete cleanup and saved_caches folders

### LINKED ISSUE #26: clean_path fails with mounted file systems
The clean path command fails if the folder to be wiped is a docker volume mount

Changing
- subprocess.check_output(['sudo', '-u', p.owner(), 'rm', '-rf', str(p)])
to
- subprocess.check_output(['sudo', '-u', p.owner(), 'rm', '-rf', str(p) + '/*'])
should allow the folder to be cleared successfully

arodrime:
Fixes #26, #109 an #164.
Fixes master bug in config: https://github.com/thelastpickle/cassandra-medusa/pull/168/files#diff-3feb4fc181a6e98475e3a0899f71011f
Improve tests (so they show the error better): https://github.com/thelastpickle/cassandra-medusa/pull/168/files#diff-dab28e770b45220deae663cad43f2227

### DIFF
--- a/medusa/restore_node.py
+++ b/medusa/restore_node.py
@@ -86,8 +86,8 @@ def restore_node_locally(config, temp_dir, backup_name, in_place, keep_auth, see
 
     # Clean the commitlogs, the saved cache to prevent any kind of conflict
     # especially around system tables.
-    clean_path(cassandra.commit_logs_path)
-    clean_path(cassandra.saved_caches_path)
+    clean_path(cassandra.commit_logs_path, keep_folder=True)
+    clean_path(cassandra.saved_caches_path, keep_folder=True)
 
     # move backup data to Cassandra data directory according to system table
     logging.info('Moving backup data to Cassandra data directory')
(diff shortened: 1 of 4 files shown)


arodrime:
Test pass now, smoke test looks ok. Doing some more test but this can be reviewed already I think @adejanovski.

medusa/restore_node.py
```diff
@@ -225,10 +228,19 @@ def table_is_allowed_to_restore(keyspace, table, fqtns_to_restore):
...
+def clean_path(p, keep_folder=False):
+    path = str(p)
+    if p.exists() and os.path.isdir(path) and len(os.listdir(path)):
+        logging.debug('Cleaning ({})'.format(path))
+        if keep_folder:
+            logging.debug('Removing files - keep folder {}'.format(path))
+            for f in os.listdir(path):
+                file_path = os.path.join(path, f)
+                logging.debug('Removing file {}'.format(file_path))
+                os.remove(file_path)
```
adejanovski, line 240: Why not use the `rm` command with subprocess here as well? Or the other way around so we're consistent in how we access the FS. Either all through Python modules (more portable code) or with shell commands (easier to deal with file ownership?).

arodrime: Makes sense, I thought about that, and dropped the ball. 100% agree on making this consistent, and we use bash commands everywhere else 👍.

arodrime: Here is the diff, I'll make sure this works.

```
-                os.remove(file_path)
+                subprocess.check_output(['sudo', '-u', p.owner(), 'rm', '-rf', file_path])
```

Issue #176, opened 2020-07-24
TLS1.0 is deprecated in most places


mkramer07:
Really just want to upgrade the tls version that's pulled from ssl to now use 1.2 instead of 1.0. Most places have already migrated to 1.2 and not even sure how many people are using 1.0 these days.

mkramer07:
It's a pretty quick and easy fix, I have running locally with just changing `PROTOCOL_TLSv1` to `PROTOCOL_TLSv1_2`

arodrime:
@mkramer07 feel free to open a PR if you'd like. Either ways, I'll check it as soon as possible. Thanks for the feedback.

ernstae:
I'd like to adopt Medusa, but TLS 1.2 is non-negotiable in my world.  Any movement on this feature?

adejanovski:
Yes, we're working on this in conjunction with some work that is being done on Cassandra 4.0.
We should update Medusa soon to support a different version of TLS.

mkcello96:
HI all, I put in a PR here https://github.com/thelastpickle/cassandra-medusa/pull/260 (I work with @ernstae). Ready for review, hopefully can just be a replacement for the work you have slated to be done.  
@arodrime @adejanovski @mkramer07

PR #177, opened 2020-07-24
Upgrading TLS version

mkramer07:
fixes https://github.com/thelastpickle/cassandra-medusa/issues/176

Simple upgrade of TLS version


arodrime:
@mkramer07 what about using `PROTOCOL_TLS` instead, to let the lib pick the latest supported version: https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS ? 

It seems that even `PROTOCOL_TLSv1_2` is deprecated somehow: https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1_2

mkcello96:
I put together a PR that follows @arodrime's suggestions here: https://github.com/thelastpickle/cassandra-medusa/pull/260

adejanovski:
TLS update was merged in #277.
Thanks a lot @mkramer07 for starting the conversation and pushing this PR!

PR #189, opened 2020-08-11
Add s3 canned ACL support

ivanmp91:
Changing default canned ACL[1] would be useful for situations where the S3 bucket is located in a different account. For S3, by default the object owner is the source AWS account where the object comes from [2]. This is a big limitation for a disaster recovery scenario where the restore could be done in a different account and requires to previously change the permissions to all objects in the bucket.

The default ACL value will remain as private.

[1] https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
[2] https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-owner-access/

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1406) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1406
┆priority: Medium

### DIFF
--- a/medusa/storage/aws_s3_storage/awscli.py
+++ b/medusa/storage/aws_s3_storage/awscli.py
@@ -27,6 +27,7 @@ class AwsCli(object):
     def __init__(self, storage):
         self._config = storage.config
         self.storage = storage
+        self.canned_acl = storage.config.canned_acl
 
     @property
     def bucket_name(self):
@@ -73,7 +74,8 @@ class AwsCli(object):
         awscli_output = "/tmp/awscli_{0}.output".format(job_id)
         objects = []
         for src in srcs:
-            cmd = [self._aws_cli_path, "s3", "cp", str(src), "s3://{}/{}".format(bucket_name, dest)]
+            cmd = [self._aws_cli_path, "s3", "cp", "--acl", self.canned_acl,
+                   str(src), "s3://{}/{}".format(bucket_name, dest)]
             objects.append(self.upload_file(cmd, dest, awscli_output))
 
         return objects
--- a/docs/aws_s3_setup.md
+++ b/docs/aws_s3_setup.md
@@ -51,6 +51,7 @@ Create an IAM Policy called `MedusaStorageStrategy`, with the following definiti
                 "s3:GetReplicationConfiguration",
                 "s3:ListMultipartUploadParts",
                 "s3:PutObject",
+                "s3:PutObjectAcl",
                 "s3:GetObject",
                 "s3:GetObjectTorrent",
                 "s3:PutObjectRetention",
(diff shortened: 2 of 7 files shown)


adejanovski:
Hi @ivanmp91,

we had a small git problem which made us rewrite history in `master`...
Could you please reset your master branch and rebase this branch on top of it?

Thanks and sorry for the inconvenience!

ivanmp91:
> Hi @ivanmp91,
> 
> we had a small git problem which made us rewrite history in `master`...
> Could you please reset your master branch and rebase this branch on top of it?
> 
> Thanks and sorry for the inconvenience!

Sure! This is done @adejanovski

adejanovski:
Hi @ivanmp91,

there are a lot of commits in this PR that would need squashing and cleaning up.
Could you rebase on top of the latest `master`, drop the commits that are part of this PR but were not written by you, and squash the remaining commits into a single one?
I can do this for you if you're not at ease with this type of manipulations with git.

medusa/storage/aws_s3_storage/awscli.py
```diff
@@ -73,7 +74,8 @@ def cp_upload(self, *, srcs, bucket_name, dest, max_retries=5):
         awscli_output = "/tmp/awscli_{0}.output".format(job_id)
         objects = []
         for src in srcs:
-            cmd = [self._aws_cli_path, "s3", "cp", str(src), "s3://{}/{}".format(bucket_name, dest)]
+            cmd = [self._aws_cli_path, "s3", "cp", "--acl", self.canned_acl,
```
adejanovski, line 77: Would this be backwards compatible with the current behavior if set to `private` by default?
I seem to understand that it would, but want to make sure of it.

adejanovski: Also, we're moving to a consolidated set of classes for all S3 compatible backends in #272. I'm wondering how this would affect other storage providers than AWS.
I'd be in favor of making this flag added to the aws command line only if `canned_acl` is set to anything else but `private` (or maybe we could leave it unset by default).  
@burmanm, wdyt?

burmanm: Yeah, I only added parameters to the command line if they differ from the default.

ivanmp91: @adejanovski according to aws s3 documentation, private is the default ACL (https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl) which keeping it as a default option if is not set shouldn't change the behavior.

adejanovski:
ouch there are some conflicts we need to handle before it can be merged.

PR #192, opened 2020-08-25
Relative installation path to config example

dmitry-saprykin:
Absolute path makes it impossible to install into virtual environment without sudo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,5 +64,5 @@ setuptools.setup(
             'medusa=medusa.medusacli:cli',
         ]},
     scripts=['bin/medusa-wrapper'],
-    data_files=[('/etc/medusa', ['medusa-example.ini'])]
+    data_files=[('etc', ['medusa-example.ini'])]
 )


setup.py
```diff
@@ -64,5 +64,5 @@
             'medusa=medusa.medusacli:cli',
         ]},
     scripts=['bin/medusa-wrapper'],
-    data_files=[('/etc/medusa', ['medusa-example.ini'])]
+    data_files=[('etc', ['medusa-example.ini'])]
```
adejanovski, line 67: This is actually being used during the installation phase to place the example in file into `/etc/medusa`. 
What's the behavior with this change?

adejanovski (commented):
I'm not sure to understand what this change does and why it's necessary.
Any additional information would be appreciated.

setup.py
dmitry-saprykin: Hi @adejanovski ,

I think I got into python issue not medusa. I had an issue trying to install medusa wheel.
pip was installing medusa-example.ini into absolute /etc/medusa path even for virtual environment.
I tried to reproduce it today and medusa-example.ini was installed into site-packages as expected.

julienlau:
Please note that you also have the --root option to install without sudo:
`python3 setup.py install --prefix=. --root=/opt/cassandra-medusa`

PR #199, opened 2020-09-02
Remove nc (netcat) dependency, use python socket instead

### LINKED ISSUE #198: Remove use of OS function 'nc' (netcat) and use a python equivalent instead
Using `nc` command, we created a dependency at the OS level that it is hard to fix other than by adding pre-requisite.

Instead we could use a python version of netcat, doing the same but that we can add to medusa code with something like:

```
import socket
def netcat(hostname, port, content):
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.connect((hostname, port))
    s.sendall(content)
    s.shutdown(socket.SHUT_WR)
    while 1:
        data = s.recv(1024)
        if data == "":
            break
        print "Received:", repr(data)
    print "Connection closed."
    s.close()
```

From: https://stackoverflow.com/questions/1908878/netcat-implementation-in-python#answer-1909355
 (or use an external python library, that we can add to dependencies).

This aims at solving problems as referenced in #187:
> On first run I noticed you seem to require netcat but is not mentioned. It might be good to mak the error or requisite clear, since many containers, instances, etc are using reduced OSes which don't include it :)
I installed it across nodes to fix the error but doesn't seem related since scylla start error persists:
stderr:timeout: failed to run command ‘nc’: No such file or directory

And other similar feedbacks we received.

arodrime:
Fixes #198

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -29,7 +29,7 @@ import uuid
 import yaml
 
 from subprocess import PIPE
-
+from retrying import retry
 from cassandra.cluster import Cluster, ExecutionProfile
 from cassandra.policies import WhiteListRoundRobinPolicy
 from cassandra.auth import PlainTextAuthProvider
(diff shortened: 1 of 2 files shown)


medusa/cassandra_utils.py
```diff
@@ -640,15 +638,35 @@ def is_ccm_up(args, nodetool_command):
...
+        is_accessible = False
+    finally:
+        s.close()
+        return is_accessible
+
+
+def is_cassandra_up(host, port):
+    host_up = False
+    retry = 5
+    for i in range(retry):
```
rzvoncek, line 661: We are using the retrying module elsewhere in Medusa code. Perhaps we can use it here too?

rzvoncek (changes requested):
I'm not feeling strongly about the change, but please consider using the retrying module.

medusa/cassandra_utils.py
arodrime: @rzvoncek sounds reasonable. Please would you check the last commit.

I will also squash before merge, once you could review the changes of the last commit. The intent was to make changes clear for you to review.

rzvoncek (approved):

arodrime:
Just squashed the commits. Considering the review is still valid. Merging now.

PR #205, opened 2020-09-23
Fix GCS folders being uploaded as files when they only contain the schema.cql file

adejanovski:
Empty tables get uploaded to GCS in a weird way as they appear as a file instead of a directory, whose content is the content of the schema.cql file 🙄 

It seems like in some edge cases, this breaks the verification of backups which breaks them.
This PR skips the schema.cql (the one generated by Cassandra during a snapshot) which then allows to skip empty tables altogether.


arodrime:
@adejanovski this was part of #209, closing this one. If I'm wrong please reopen.

PR #209, opened 2020-09-30
Enable multi Cassandra version integration tests

adejanovski:


arodrime:
@adejanovski  I think I preferred my option (excluding experimental flag tests for other than trunk, and excluding non-experimental legs for the trunk).

Here S3/GCP on trunk are 'non-experimental'. Happy to talk about this but I think I had it nailed. Ok to add the python 3.8 to the matrice of course.

adejanovski:
> @adejanovski I think I preferred my option (excluding experimental flag tests for other than trunk, and excluding non-experimental legs for the trunk).
> 
> Here S3/GCP on trunk are 'non-experimental'. Happy to talk about this but I think I had it nailed. Ok to add the python 3.8 to the matrice of course.

Actually, S3 and GCP are passing with trunk because they're not using encryption. They should not be experimental as failure would mean something broke.

PR #211, opened 2020-10-12
Honor concurrent_transfers in GCP storage. Fixes #103

### LINKED ISSUE #103: GCS uploads won't comply with the concurrent_transfers setting
It's currently not possible to control concurrency on GCS uploads as the `concurrent_transfers` setting is ignored with this storage backend.
We could use the `-o "GSUtil:parallel_process_count=X"   -o "GSUtil:parallel_thread_count=X"` arguments to have gsutil use the `concurrent_transfers` setting properly.

rzvoncek:


arodrime (commented):
I had a look. If I remember well I tried to set up those then removed it (that’s why options were commented here: https://github.com/thelastpickle/cassandra-medusa/blob/a5df7c5f54e83f609c4c691aaf195738e91b353d/medusa/storage/google_cloud_storage/gsutil.py#L69-L74).

The problem is that I don’t remember why I removed it... (oops).

Also I’m not sure to understand this comment here: https://github.com/thelastpickle/cassandra-medusa/blob/a5df7c5f54e83f609c4c691aaf195738e91b353d/medusa/storage/google_cloud_storage/gsutil.py#L76-L77

But it seems we will have 1 process, but with 5 (or 24)  threads. Is that as intended @rzvoncek? Shouldn't we also limit the threads to 1 by default? 🤔 

Other than this I'm completely happy with fixing concurrent processes by default to 1 to ensure we don't harm as much as we can.

arodrime:
Another question is whether you want this merged in master (and next 0.8.x releases only) or merge in 0.7 (0.7.2 should be out soon).

rzvoncek:
For gsutil, there are two knobs we can turn - one controls the number of processes gsutil spawns, the second one controls the number of threads inside those processes. By default, there's as many processes as there are cores, with 5 threads in each.

I think we only want to control the number of processes. That's what we do for s3. We want to keep the threads to default, so if the process runs, it still has some room to work well. I think it's a good middle ground.

I'm fine with just a merge to master, no need for a release. I need this in a context where Medusa gets packaged separately, straight off the master branch.

arodrime (approved):
Looks good to me.

Issue #201, opened 2020-09-07
Medusa installed with apt produces error message


markkutikkanen:
I didn't want to go configuring python environment and dependencies manually so I was happy to see you provide an APT repo. https://github.com/thelastpickle/cassandra-medusa/wiki/Installation#debian-packages

I followed the instructions and there was no errors.

However after installing any `medusa` invocation produces error:
```
~$ medusa
Traceback (most recent call last):
  File "/usr/bin/medusa", line 6, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named 'pkg_resources'
```

The environment is Ubuntu 16.04.3

arodrime:
Being discussed on slack here: https://the-asf.slack.com/archives/C0111B5KADA/p1599476136004200

vinovarov:
Exactly same error on Debian 9.13, no googled workaround like re-installing python setuptools helped.

Would be great to have a working deb package as we would have to deploy Medusa (if we'll be able to adopt it).

vinovarov:
Fix from https://github.com/thelastpickle/cassandra-medusa/issues/208 seems to work for apt-installed Medusa too, at least for me:

```
apt install python3-pip -y
pip3 install ssh2-python==0.19.0
```

Please confirm somebody.

masterzen:
Go the same issue. BTW 0.7.0 is also affected.
I believe this is an issue with the virtualenv that is packaged, but I couldn't find what was wrong or how to heal it.

masterzen:
A bit more information: the package works on a bionic 18.04 machine, but not on a xenial 16.04 machine. I think the reason is because the package is built in a 18.04 container, so assuming a python 3.8 environment.
I was able to rebuild a xenial package (but that required a few changes to the build system).

masterzen:
I've started working on it @arodrime. I'm waiting for the packages to be built to test them and if that's working find, I'll create a PR.

arodrime:
@masterzen that's amazing, thanks for your time! I'll do my best to review your request quickly once it's out and probably cut a release soon-ish!

arodrime:
Closing this issue 'as fixed' in #213. Thanks @masterzen!

PR #213, opened 2020-10-12
(#201) Ubuntu 16.04 & 18.04 debian packaging

masterzen:
The current debian package generated was not working properly on ubuntu xenial (see #201). This change allows to fix this by extending the way how the debian package is built.
`build-deb.sh` will now build a package for ubuntu 16.04 (xenial) and 18.04 (bionic) 
Both packages version reflect the distribution codename they've been built for, so they can be deployed to the same debian repository.
It would be relatively easy to extend the package build to other ubuntu or debian releases (provided they run a relatively recent python3 and have `dh_virtualenv >= 1.0`).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+debian/cassandra-medusa/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,35 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v1
+    - name: Build and publish Debian
+      env:
+        BINTRAY_USERNAME: ${{ secrets.BINTRAY_USERNAME }}
+        BINTRAY_KEY: ${{ secrets.BINTRAY_KEY }}
+        SUITES: "bionic xenial buster stretch"
+      run: |
+        cd packaging/docker-build
+        $all_suites=($SUITES) # dash supports arrays expansion
+        for suite in $all_suites
+        do
+          docker-compose build "cassandra-medusa-builder-${suite} \
+            && docker-compose run "cassandra-medusa-builder-${suite}"
+        done
+        cd ../..
+        version=$(cat VERSION)
+        for suite in $all_suites
+        do
+          if [ -f "packages/cassandra-medusa_${version}-0~${suite}0_amd64.deb" ]; then
+            echo "uploading packages/cassandra-medusa_${version}-0~${suite}0_amd64.deb from ${suite} to Bintray"
+            curl -T "packages/cassandra-medusa_${version}-0~${suite}0_amd64.deb" \
+              -u${BINTRAY_USERNAME}:${BINTRAY_KEY} \
+              -H X-Bintray-Debian-Distribution:${suite} -H X-Bintray-Debian-Component:main -H X-Bintray-Debian-Architecture:amd64 -H X-Bintray-Version:${version} \
+              https://api.bintray.com/content/thelastpickle/medusa-deb/cassandra-medusa/${version}/cassandra-medusa_${version}-0~${suite}0_amd64.deb
+          else
+            echo "Error: no packages found for ${suite}"
+          fi
+        done
+        echo "Publishing release in Bintray"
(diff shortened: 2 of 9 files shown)


adejanovski:
Thanks @masterzen! Reviewing this PR.

adejanovski (commented):
Hi @masterzen,

packages are correctly built for both Ubuntu versions with your changes, but it will break our release process if the Debian publishing step isn't adjusted appropriately: https://github.com/thelastpickle/cassandra-medusa/blob/master/.github/workflows/ci.yml#L164-L180

I guess we should upload both files in different distribution specific folders, going from `/cassandra-medusa/${version}/cassandra-medusa_${version}_amd64.deb` to `/cassandra-medusa/${version}/${distribution}/cassandra-medusa_${version}-0~${distribution}0_amd64.deb` and adjust the distribution header appropriately (currently `-H X-Bintray-Debian-Distribution:wheezy,bionic,jessie,xenial`).

I'd also like to swap debian and Pypi uploads because the latter doesn't allow re-uploading an existing version, and if we don't get the debian uploads right on the first attempt, then we'll be stuck with upgrading the version number and releasing again.

Let me know what you think.

masterzen:
> Hi @masterzen,
> 
> packages are correctly built for both Ubuntu versions with your changes, but it will break our release process if the Debian publishing step isn't adjusted appropriately: [`master`/.github/workflows/ci.yml#L164-L180](https://github.com/thelastpickle/cassandra-medusa/blob/master/.github/workflows/ci.yml#L164-L180)

Indeed, I didn't think about checking the release process.
This means I need to also change to build packages for debian versions. 
I see you were targeting wheezy and jessie which are very old versions. The current stable and oldstable are buster and stretch, which IMHO would be better to support.

So we could build a distro specific package for:
* xenial
* bionic
* stretch
* buster

And probably latest ubuntu LTS 20.04 (feral). 

> I guess we should upload both files in different distribution specific folders, going from `/cassandra-medusa/${version}/cassandra-medusa_${version}_amd64.deb` to `/cassandra-medusa/${version}/${distribution}/cassandra-medusa_${version}-0~${distribution}0_amd64.deb` and adjust the distribution header appropriately (currently `-H X-Bintray-Debian-Distribution:wheezy,bionic,jessie,xenial`).
> 

I chose to add the distribution to the version so that you could keep the same folder layout in the repository. The only thing that should be done is to upload each package separately in their correct distribution (ie `-X X-Bintray-Debian-Distribution: xenial` for the xenial package, etc).

There won't be any naming collision in the repository pool.

> I'd also like to swap debian and Pypi uploads because the latter doesn't allow re-uploading an existing version, and if we don't get the debian uploads right on the first attempt, then we'll be stuck with upgrading the version number and releasing again.

I see. That's indeed a good idea 👍 I

> Let me know what you think.

That sounds like a good plan. I'm going to see if I can work this week on adding debian distribution packaging (following the same model as I did for ubuntu) and contribute the worklow step.

Thanks for having a look to this PR :)

masterzen:
@adejanovski,

I've pushed a new revision that:
* also builds buster and stretch packages (along with bionic and xenial)
* modifies the package upload to bintray as discussed.

I wasn't able to test the bintray upload, so it might not work at all :)

Let me know if you have any comments or questions,
Thanks !

adejanovski (approved):
I have just one minor change that would be awesome to have in `build-deb.sh` and aside from could you squash your commits into a single one?

For the publishing, I guess we'll find out soon enough 😅 I'll be running some tests manually with this code to see if everything works as expected before we issue a release.

Thanks!

masterzen:
> I have just one minor change that would be awesome to have in `build-deb.sh` and aside from could you squash your commits into a single one?

Done & done.

> For the publishing, I guess we'll find out soon enough 😅 I'll be running some tests manually with this code to see if everything works as expected before we issue a release.

I hope this will work fine, feel free to ping me if there are any issues to fix.

adejanovski:
Awesome! 
Thanks for this fine contribution and for the quick interactions @masterzen 🙏 

Merging the PR.

PR #215, opened 2020-10-20
Restrict allowed gevent versions

adejanovski:
The gevent dependency above 1.4.0 seems to break builds in some environments.
Let's limit the allowed versions to prevent that kind of breakages.


arodrime (commented):
Change looks good, yet you might want to add the same line in setup.py file: https://github.com/thelastpickle/cassandra-medusa/blob/master/setup.py#L42-L58

Also there is Python 3.5 mentioned in setup.py, let's move it to Python 3.6?
https://github.com/thelastpickle/cassandra-medusa/blob/master/setup.py#L38-L42

adejanovski:
Turns out to be unnecessary.

Closing.

PR #218, opened 2020-10-22
[gcs] Use 'gsutil -m' only if concurrent_transfers > 1

rzvoncek:


adejanovski (approved):

rzvoncek:
I've rebased on master.

PR #221, opened 2020-10-23
Allow to download only a set of keyspaces/tables

masterzen:
The `medusa download` command allows only to download a whole backup.

This change adds the same filtering flags as the `restore-node` command:
--keyspace <ks>
--table <ks.ta>

This allows to specify which tables or keyspaces combinations medusa will download from the storage.

By default, `medusa download` always downloads the system keyspaces. I added a new option,
`--ignore-system-keyspaces` that allows to not download them. This is very useful if you just want to download a specific keyspace or table.

I noticed that `medusa download` always creates unnecessary local directories (one per keyspace). When downloading only some keyspaces or specific tables, there's no reason all the other keyspaces directory needs to be created. I modified this behavior to only create the downloaded keyspaces directories.

### DIFF
--- a/medusa/filtering.py
+++ b/medusa/filtering.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020- Datastax, Inc. All rights reserved.
+# Copyright 2019 Spotify AB. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+
+
+def filter_fqtns(keep_keyspaces, keep_tables, manifest, ignore_system_keyspaces=False):
+    retained = set()
+    ignored = set()
+    manifest = json.loads(manifest)
+
+    for section in manifest:
+        ks = section['keyspace']
+        # in manifest, the table names have cfids, but from CLI we get it without
+        # we need to take care and use both
+        t = section['columnfamily'].split('-')[0]
+
+        fqtn = '{}.{}'.format(ks, t)
+        fqtn_with_id = '{}.{}'.format(section['keyspace'], section['columnfamily'])
+
+        # if not keyspaces / tables were specified, we keep everything
+        if len(keep_keyspaces) == 0 and len(keep_tables) == 0:
+            retained.add(fqtn_with_id)
+            continue
+
+        # if the whole keyspace is a keep, or a system keyspace (C* internal)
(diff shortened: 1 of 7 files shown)


masterzen:
I fixed the integration test failure, except the last one for which I'm a bit puzzled.

Note that the gcs and s3 test variant don't do anything in PRs (I suppose it's made to limit cost).

medusa/filtering.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Spotify AB. All rights reserved.
```
adejanovski, line 2: @michaelsembwever, which copyright should be added here?
Definitely not the Spotify one, but I'm wondering what's our approach for external contributions.

I like the approach described [here](https://www.linuxfoundation.org/blog/2020/01/copyright-notices-in-open-source-software-projects/), under "Community best practices". 
wdyt?

medusa/filtering.py
```diff
@@ -0,0 +1,52 @@
...
+        fqtn = '{}.{}'.format(ks, t)
+        fqtn_with_id = '{}.{}'.format(section['keyspace'], section['columnfamily'])
+
+        # if not keyspaces / tables were specified, we keep everything
+        if len(keep_keyspaces) == 0 and len(keep_tables) == 0:
+            retained.add(fqtn_with_id)
+            continue
+
+        # if the whole keyspace is a keep, or a system keyspace (C* internal)
+        if ks in keep_keyspaces or (not ignore_system_keyspaces and (ks in SYSTEM_KEYSPACES)):
```
adejanovski, line 41: Would you mind extracting this condition into a separate method? It's a little hard to read.

adejanovski (changes requested):
LGTM aside from a minor comment on the code and some questions we need to answer on copyrights.

adejanovski:
Hi @masterzen,

Let me know if you need help moving this PR forward.

Thanks!

masterzen:
@adejanovski, sorry I was quite busy those last weeks. I'll hope to be able to fix the issues you pointed in your comments ASAP.

BTW, did you had an answer about the copyright ?

I'm personally fine with the approach you linked, using something like "Copyright Contributors to the Cassandra Medusa project"

Also, I need to understand why the last integration tests doesn't pass. It apparently errors when calling `sstableloader`.

And finally, I'm not sure it is normal, but the GCS and S3 integration tests never do anything in the PR.

adejanovski:
Sure thing, no worries.

For the copyright, would you mind adding the Datastax copyright?

```
# Copyright 2020 DataStax Inc.
```

You still hold copyright to your contribution, which is tracked by GitHub, but gives Datastax permission to donate the code to the ASF for example or legally defend/protect it.

The failing integration tests have been fixed in `master` and in Cassandra itself, there's no relation to your PR.
The GCS and S3 integration tests use GitHub Actions secrets to get credentials for GCP and AWS, which are not available to PRs coming from external contribution for security reasons.
These tests will always be skipped in this case, and that's an expected behavior.

adejanovski:
You'll need to rebase anyway due to conflicts on `medusa/download.py`, which will make your PR pass all integration tests.

masterzen:
@adejanovski, I've pushed a new version that took into account your comments.

Regarding the copyright, I left the Spotify one, because most of that code was extracted from a file that had this copyright.

Let me know if you need anything else.
Thanks !

adejanovski (approved):

PR #223, opened 2020-11-02
gRPC service

jsanda:
This PR is for #137 and is in place of #141.


VERSION
```diff
@@ -1 +1 @@
-0.8.0-dev
+0.7.1-dev
```
adejanovski, line 1: You'll need a little rebase before merging.

medusa/config.py
```diff
@@ -122,6 +128,11 @@ def load_config(args, config_file):
         'monitoring_provider': 'None'
     }
 
+    config['grpc'] = {
+        'enabled': False,
+        'cassandra_url': 'http://localhost:8778/jolokia/'
```
adejanovski, line 133: Could the value be moved to a constant?

medusa/restore_node.py
```diff
@@ -74,21 +75,23 @@ def restore_node_locally(config, temp_dir, backup_name, in_place, keep_auth, see
...
     logging.info('Downloading data from backup to {}'.format(download_dir))
     download_data(config.storage, node_backup, fqtns_to_restore, destination=download_dir)
 
-    logging.info('Stopping Cassandra')
-    cassandra.shutdown()
-    wait_for_node_to_go_down(config, cassandra.hostname)
+    if not medusa.utils.evaluate_boolean(config.grpc.enabled):
+        logging.info('Stopping Cassandra')
+        cassandra.shutdown()
+        wait_for_node_to_go_down(config, cassandra.hostname)
```
adejanovski, line 88: This means the gRPC server can exclusively be used for k8s clusters? I thought we would have the option to use it for classic clusters as well and ease automation for folks who prefer invoking Medusa remotely.
Maybe we should have another config setting to indicate whether or not we're dealing with k8s.

medusa/restore_node.py
```diff
@@ -74,21 +75,23 @@ def restore_node_locally(config, temp_dir, backup_name, in_place, keep_auth, see
...
 
     # Clean the commitlogs, the saved cache to prevent any kind of conflict
     # especially around system tables.
-    clean_path(cassandra.commit_logs_path, keep_folder=True)
-    clean_path(cassandra.saved_caches_path, keep_folder=True)
+    use_sudo = not medusa.utils.evaluate_boolean(config.grpc.enabled)
```
adejanovski, line 92: Same here, this should be subject to running into k8s or not instead of grpc or not.

medusa/storage/google_storage.py
```diff
@@ -30,6 +30,9 @@
 from medusa.storage.google_cloud_storage.gsutil import GSUtil
 
 
+GSUTIL_MAX_FILES_PER_CHUNK = 64
```
adejanovski, line 33: This doesn't belong in this PR. Either we have a missing commit in `master` or this PR should be properly rebased against master.
I'll check what's wrong here (and the rebase is needed anyway).

setup.py
```diff
@@ -58,7 +58,8 @@
     ],
     extras_require={
         'S3': ["awscli>=1.16.291"],
-        'GCS': ["google-cloud-storage>=1.7.0"]
+        'GCS': ["google-cloud-storage>=1.7.0"],
+        'GRPC': ["protobuf>=3.12.0", "grpcio>=1.29.0", "grpcio-health-checking-1.29.0", "grpcio-tools>=1.29.0"]
```
adejanovski, line 62: Since the code is there and it is imported, won't the grpc requirements be needed in all scenarios?

adejanovski (approved):
I have a few comments around the fact that the grpc server seems to be exclusively available for k8s while users could benefit from it even on non dockerized environments.
I'll let @arodrime run some tests but I'm ok with merging the PR if it gets properly rebased against master and commits get squashed into a single one.

medusa/storage/google_storage.py
arodrime: @jsanda - Alex understood where this was coming from and fixed it. Rebase should do now that we updated master.

medusa/restore_node.py
jsanda: > I thought we would have the option to use it for classic clusters as well and ease automation for folks who prefer invoking Medusa remotely.

Absolutely. There is nothing Kubernetes specific about the gRPC service. There are integration tests that exercise it, and they do not run in Kubernetes.

> Maybe we should have another config setting to indicate whether or not we're dealing with k8s.

This would be my preference as well. I will add another config setting.

emerkle826: I've added a K8s configuration setting now and changed the logic to trigger based on that.

medusa/config.py
emerkle826: done

medusa/restore_node.py
emerkle826: done

setup.py
emerkle826: Indeed. Moved them to install_requires

emerkle826:
@adejanovski I would appreciate a re-review of this. I believe I have addressed all the comments so far.

medusa/service/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Spotify AB. All rights reserved.
```
adejanovski, line 2: Wrong copyright

medusa/service/grpc/__init__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Spotify AB. All rights reserved.
```
adejanovski, line 2: Wrong copyright

medusa/service/grpc/client.py
```diff
@@ -0,0 +1,50 @@
+import grpc
```
adejanovski, line 1: Missing copyright/license

medusa/service/grpc/restore.py
```diff
@@ -0,0 +1,58 @@
+import logging
```
adejanovski, line 1: Missing copyright/license

medusa/service/grpc/server.py
```diff
@@ -0,0 +1,152 @@
+import logging
```
adejanovski, line 1: Missing copyright/license

medusa/utils.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Spotify AB. All rights reserved.
```
adejanovski, line 2: Wrong copyright

adejanovski (changes requested):
@jsanda @emerkle826,

we're very close to merging this PR. There's still a code path which is specific to grpc while it should be specific to k8s instead on taking snapshots with Jolokia.
Also there are some wrong copyrights (should be DataStax Inc. instead of Spotify) and some missing license headers on new files.

adejanovski:
@emerkle826, could you squash your commits into a single one once the requested changes are done?
Thanks!

emerkle826:
> @emerkle826, could you squash your commits into a single one once the requested changes are done?
> Thanks!

Will do.

medusa/cassandra_utils.py
```diff
@@ -410,30 +416,64 @@ def create_snapshot(self, backup_name):
...
         tag = "{}{}".format(self.SNAPSHOT_PREFIX, backup_name)
         if not self.snapshot_exists(tag):
-            if self._is_ccm == 1:
-                os.popen(cmd).read()
+
+            # TODO introduce abstraction layer/interface for invoking Cassandra
+            # Eventually I think we will want to introduce an abstraction layer for Cassandra's
+            # API that Medusa requires. There should be an implementation for using nodetool,
+            # one for Jolokia, and a 3rd for the management sidecard used by Cass Operator.
+            if medusa.utils.evaluate_boolean(self.grpc_config.enabled):
```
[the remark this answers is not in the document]
jsanda: That is true. We could run the gRPC service along side Cassandra where Medusa has the ability to execute `nodetool`. I agree that this should be a k8s check here.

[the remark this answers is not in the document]
emerkle826: So, should I just move the jolokia URL to the Kubernetes config, and leave the gRPC config as simply enabled or not? If the issue here is to decouple Jolokia from gRPC, then it seems the Jolokia configuration belongs with the Kubernetes config, not gRPC config.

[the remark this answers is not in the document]
adejanovski: Yes, reliance on Jolokia is a requirement for k8s deployments only as Medusa runs in a different container than Cassandra. But we can use gRPC in a non k8s environment where we can run nodetool commands to take snapshots.

emerkle826:
@adejanovski Thanks for the reviews again. I've squashed everything down now. The 2 failing integration tests I believe are due to a bug in the latest Cassandra trunk with SSLContext initialization. I created a ticket here:

https://issues.apache.org/jira/browse/CASSANDRA-16362

Not sure when that will get addressed.

jsanda:
I assume those failures are not specific to changes in this PR? If that is correct, will we be able to merge with those failures?

adejanovski:
@jsanda @emerkle826,

I've pushed a fix for the backup-cluster operation which is not covered (yet) by integration tests and was broken due to a change from this PR.
The failing tests are unrelated to this PR and will not block the merge.

adejanovski (approved):

PR #225, opened 2020-11-12
debian: Fix build with pbuilder(-dist)

ecsv:
Beside various other flaws in this packaging strategy, it was not possible to build the package in a clean environment. Usually Debian packages are build with sbuild (on the official debian servers or enthusiast) or the much easier to use pbuilder(-dist) - and not some kind of unclean docker container.

These changes are now adjusting the debian build rules to allow a build via:

    # only the first time:
    pbuilder-dist buster amd64 create
    
    # (optional) regularly install newest updates in the builder environment
    find "$HOME"/pbuilder/buster-base.tgz -mmin +1440 -exec pbuilder-dist buster amd64 update \;
    
    # create .dsc - here from within the actual source directory
    debuild -S
    
    # or when the source directory is clean:
    dpkg-source -b .
    
    # start the actual build (which unfortunately needs network access - which is usually not allowed for Debian package builds)
    pbuilder-dist buster amd64 build --use-network yes ../cassandra-medusa_0.7.1.dsc

The result will then (with the default config) end up in ~/pbuilder/buster_result

The docker-build scripts were not touched. But I would highly recommend not to continue to build these packages in unclean environments

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,6 +8,7 @@ Build-Depends:
   python3-all,
   python3-all-dev,
   python3-setuptools,
+  python3-venv,
   dh-virtualenv (>= 1.0),
   build-essential
 Standards-Version: 3.9.5
--- a/debian/control
+++ b/debian/control
@@ -10,6 +10,7 @@ Build-Depends:
   python3-setuptools,
   python3-venv,
   dh-virtualenv (>= 1.0),
+  libevent-dev,
   build-essential
 Standards-Version: 3.9.5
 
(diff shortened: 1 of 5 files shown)


adejanovski:
Thanks @ecsv !

@masterzen, since you've worked on the original PR, I'd greatly appreciate your feedback.
Thanks

masterzen:
@adejanovski, 
I don't have an extensive knowledge of python package building on debian :)
I think this PR is good and fixes some issues when building directly a package.
I haven't checked what it means for the docker build version which you are using for releases, but this probably can be fixed in another PR.

@ecsv, I'm not the original author of the docker build version. I believe the intent is to be able to cross build the packages on different machines (for instance I can build the package on different ubuntu or releases while using a mac). On the contrary `debuild`, `pbuilder` etc are only available on debian and thus limits the way the packages can be built.

adejanovski (approved):

adejanovski:
Merged as part of #240

PR #227, opened 2020-11-17
Fix failing 4.0 with encryption integration tests

adejanovski:
Previous versions of Cassandra were more permissive with using the sstableloader without specifying the cipher suites nor the cassandra.yaml location (which allowed to retrieve the cipher suites).

### DIFF
--- a/medusa/restore_node.py
+++ b/medusa/restore_node.py
@@ -179,6 +179,7 @@ def invoke_sstableloader(config, download_dir, keep_auth, fqtns_to_restore, stor
                                           "github:apache/", "githubCOLONapacheSLASH"),
                                           '-d', hostname_resolver.resolve_fqdn() if cassandra_is_ccm == 0
                                           else '127.0.0.1',
+                                          '--conf-path', config.cassandra.config_file,
                                           '--username', cql_username,
                                           '--password', cql_password,
                                           '--no-progress',


medusa/restore_node.py
```diff
@@ -179,6 +179,7 @@ def invoke_sstableloader(config, download_dir, keep_auth, fqtns_to_restore, stor
                                           "github:apache/", "githubCOLONapacheSLASH"),
                                           '-d', hostname_resolver.resolve_fqdn() if cassandra_is_ccm == 0
                                           else '127.0.0.1',
+                                          '--conf-path', config.cassandra.config_file,
```
arodrime, line 182: @adejanovski shouldn't this change be made also in the case we are not using ccm?

This is enough to fix the sstable loader with 4.0? wow :)

adejanovski: This code path is used for all cases, ccm and plain old Cassandra cluster, unless I'm missing something.

arodrime: Oh yes indeed, you're right. I have been confused by the previous line. My bad 👍 .

adejanovski:
@arodrime, all tests are passing now that [CASSANDRA-16280](https://issues.apache.org/jira/browse/CASSANDRA-16280) was committed.
We can merge if you approve the PR.

arodrime:
YES PLEASE \o/.

arodrime (approved):
Looks amazing, simple and does the job!

PR #228, opened 2020-11-17
add backup-cluster command + decouple snapshots from upload

arodrime:
Work done:

* New orchestration backups (send to all nodes) --> backup-cluster
* Change pssh_pool_size a mix of: parallel_snapshots, parallel_uploads (for backups) and parallel_restores. Not just create as many ssh commands but make them sync / blocking to enforce the parallelism to be respected
* Build 'backup-node' command as an alias of 'backup' command (add dependency to click_aliases)
* Change snapshots prefixes from '<random UUID>' to 'medusa-'
* Unify pssh_run for cluster backups and restores, created 'orchestration' module.

**Impacts**
* PSSH Pool sizes are now:
  - Blocking - previously with a pool of 10 and 500 nodes, within seconds all commands were run as pssh is not waiting for commands to finish to release one connection from the pool
  - As a consequence of the above, pools (now ie. parallelism) are as 500 by default for backup restores, thus on clusters > 500 nodes, the restore might take longer as medusa will wait for the 500 first nodes to finish (this is tunable though)
  -  Parallelism is now **1** (sequential) for backups, this is really different from the current behaviour where all nodes upload data at once. It will take longer to upload backups, with the tradeoff of a lighter impact (to no impact) to the cluster. This is tunable as well.
  - Option pssh_pool_size no longer exist (replaced by parallel_* options to remove confusion and allow a split, different values.
* `medusa backup` still works, even though for coherence reasons, we added the alias `medusa backup-node` and should (maybe) deprecate `medusa backup` that I find more confusing.

### DIFF
--- a/medusa/backup_cluster.py
+++ b/medusa/backup_cluster.py
@@ -1,0 +1,179 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020- Datastax, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import sys
+import uuid
+import datetime
+import traceback
+
+import medusa.config
+from medusa.orchestration import Orchestration
+from medusa.monitoring import Monitoring
+from medusa.cassandra_utils import CqlSessionProvider, Cassandra
+from medusa.storage import Storage
+from medusa.network.hostname_resolver import HostnameResolver
+
+
+def orchestrate(config, backup_name, stagger, mode, temp_dir, parallel_snapshots, parallel_uploads):
+    backup = None
+    monitoring = Monitoring(config=config.monitoring)
+    try:
+        backup_start_time = datetime.datetime.now()
+        if not config.storage.fqdn:
+            err_msg = "The fqdn was not provided nor calculated properly."
+            logging.error(err_msg)
+            raise Exception(err_msg)
+
+        if not temp_dir.is_dir():
+            err_msg = '{} is not a directory'.format(temp_dir)
+            logging.error(err_msg)
+            raise Exception(err_msg)
+
+        try:
+            # Try to get a backup with backup_name. If it exists then we cannot take another backup with that name
+            storage = Storage(config=config.storage)
+            cluster_backup = storage.get_cluster_backup(backup_name)
+            if cluster_backup:
+                err_msg = 'Backup named {} already exists.'.format(backup_name)
+                logging.error(err_msg)
+                raise Exception(err_msg)
+        except KeyError:
+            info_msg = 'Starting backup {}'.format(backup_name)
+            logging.info(info_msg)
+
+        backup = BackupJob(config, backup_name, stagger, mode, temp_dir, parallel_snapshots, parallel_uploads)
+        backup.execute()
+
+        backup_end_time = datetime.datetime.now()
+        backup_duration = backup_end_time - backup_start_time
+
+        logging.debug('Emitting metrics')
+
+        logging.info('Backup duration: {}'.format(backup_duration.seconds))
+        tags = ['medusa-cluster-backup', 'cluster-backup-duration', backup_name]
+        monitoring.send(tags, backup_duration.seconds)
+
+        tags = ['medusa-cluster-backup', 'cluster-backup-error', backup_name]
+        monitoring.send(tags, 0)
+
+        logging.debug('Done emitting metrics.')
+        logging.info('Backup of the cluster done.')
+
+    except Exception as e:
+        tags = ['medusa-cluster-backup', 'cluster-backup-error', backup_name]
+        monitoring.send(tags, 1)
+
+        logging.error('This error happened during the cluster backup: {}'.format(str(e)))
+        traceback.print_exc()
+
+        if backup is not None:
+            err_msg = 'Something went wrong! Attempting to clean snapshots and exit.'
+            logging.error(err_msg)
+
+            delete_snapshot_command = ' '.join(backup.cassandra.delete_snapshot_command(backup.snapshot_tag))
+            pssh_run_success_cleanup = backup.orchestration_uploads\
+                .pssh_run(backup.hosts,
+                          delete_snapshot_command,
+                          hosts_variables={})
+            if pssh_run_success_cleanup:
+                info_msg = 'All nodes successfully cleared their snapshot.'
+                logging.info(info_msg)
+            else:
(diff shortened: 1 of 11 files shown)


medusa/backup_cluster.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _create_snapshots(self):
+        # Run snapshot in parallel on all nodes,
+        pssh_run_success = self.orchestration_snapshots \
+            .pssh_run(self.hosts,
+                      self.cassandra.create_snapshot_command(self.backup_name),
+                      hosts_variables={})
+        if not pssh_run_success:
+            # we could implement a retry.
+            err_msg = 'Some nodes failed to create the snapshot. Cleaning snapshots and exiting'
```
adejanovski, line 134: I don't think that message is accurate as we're not doing cleanups. Are we?

medusa/backup_cluster.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _uplaod_backup(self):
+        backup_command = self._build_backup_cmd()
+        # Run upload in parallel or sequentially according to pssh_pool_size defined by the user
+        pssh_run_success = self.orchestration_custom.pssh_run(self.hosts,
+                                                              backup_command,
+                                                              hosts_variables={})
+        if not pssh_run_success:
+            # we could implement a retry.
+            err_msg = 'Some nodes failed to upload the backup. Cleaning snapshots and exiting'
```
adejanovski, line 148: Same question here: are we really cleaning snapshots?

medusa/medusacli.py
```diff
@@ -107,17 +108,36 @@ def cli(ctx, verbosity, without_log_timestamp, config_file, **kwargs):
...
+@click.option('--stagger', default=None, type=int, help='Drop initial backups if longer than a duration in seconds')
+@click.option('--mode', default="differential", type=click.Choice(['full', 'differential']))
+@click.option('--seed-target', help='seed of the target hosts to backup', required=True)
+@click.option('--pssh-pool-size', help="Number of concurrent ssh sessions started by pssh", default=1)
+@pass_MedusaConfig
+def backup_cluster(medusaconfig, backup_name, stagger, mode, seed_target, pssh_pool_size):
+    """
+    Backup Cassandra cluster
+    """
+    medusa.restore_cluster.orchestrate(medusaconfig,
```
adejanovski, line 135:
```suggestion
    medusa.backup_cluster.orchestrate(medusaconfig,
```

adejanovski (commented):
Looks pretty great so far!
I have a few comments already and I'd need to run actual tests on a cluster once my suggestions are applied.

medusa/backup_cluster.py
```diff
@@ -0,0 +1,165 @@
...
+        if not pssh_run_success_cleanup:
+            err_msg_cleanup = 'Some nodes failed to clear the snapshot. Cleaning snapshots manually is recommended'
+            logging.error(err_msg_cleanup)
+        sys.exit(1)
+
+
+class BackupJob(object):
+    def __init__(self, config, backup_name, stagger, mode, seed_target, pssh_pool_size):
+        self.id = uuid.uuid4()
+        self.orchestration_snapshots = Orchestration(config, 1000)
```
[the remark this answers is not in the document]
arodrime: Make sense to make it configurable. I'd keep it as high as 500 or 1000 so default is doing snapshots at once on most of the clusters:

From [doc](https://parallel-ssh.readthedocs.io/en/0.91.2/pssh_client.html):

> pool_size (int) – (Optional) Greenlet pool size. Controls on how many hosts to execute tasks in parallel. Defaults to 10. Values over 500 are not likely to increase performance due to overhead in the single thread running our event loop.

Or this [doc](https://pypi.org/project/parallel-ssh/#scaling):

>In general, long lived commands with little or no output gathering will scale better. Pool sizes in the multiple thousands have been used successfully with little CPU overhead in the single thread running them in these use cases.

That's why 500 or 1000 seem to be still is an acceptable value I guess. Also, when having less servers, less connections are opened, that's why I was considering forcing a high number there. Nonetheless I agree that if some people have troubles it's good if they can reduce this value without patching...

I will make it configurable, but I would keep 500 or 1000 as the default.

medusa/backup_cluster.py
```diff
@@ -0,0 +1,165 @@
...
+                      hosts_variables={})
+        if not pssh_run_success:
+            # we could implement a retry.
+            err_msg = 'Some nodes failed to create the snapshot. Cleaning snapshots and exiting'
+            logging.error(err_msg)
+            raise Exception(err_msg)
+
+        logging.info('A snapshot {} was created on all nodes.'.format(self.snapshot_tag))
+
+    def _uplaod_backup(self):
```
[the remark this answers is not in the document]
arodrime: oops :). Thanks.

medusa/backup_cluster.py
arodrime: That's because I moved the cleanup of the snapshot on the execute exception handle at the last moment. This comment needs to be split and the part about cleaning snapshot needs to go where we actually do it (same place where we return a metric saying backups failed). Good catch.

robbybb:
Two questions:
I see this in the design statement: Parallelism is now 1 (sequential) for backups
Are you saying there is no option in orchestration to make Parallelism > 1 or are you saying that the default is 1 and can be changed to something > 1?

Also, how robust in pssh retry?  We are backing up in the middle of the night when IBM cloud team might be messing with network connections -- typically for 5 minutes.  I don't think we need a retry outside of what pssh does if pssh is robust enough in its retries.  I'm not familiar with pssh though so I am asking.

arodrime (commented):
> Two questions:
> I see this in the design statement: Parallelism is now 1 (sequential) for backups
> Are you saying there is no option in orchestration to make Parallelism > 1 or are you saying that the default is 1 and can be changed to something > 1?

It's the default, sequentially upload backup. But you can perfectly change the command to be:

```
medusa backup-cluster --backup-name my-cluster-backup --parallel-uploads 4
```

This will induce 4 parallel uploads of the data for example.

See this here: https://github.com/thelastpickle/cassandra-medusa/pull/228/files#diff-008e28c1d3eaa50d31be537233ab33dd716ba23bca0e4c125d0ada18d625510eR132-R133

> Also, how robust in pssh retry? We are backing up in the middle of the night when IBM cloud team might be messing with network connections -- typically for 5 minutes. I don't think we need a retry outside of what pssh does if pssh is robust enough in its retries. I'm not familiar with pssh though so I am asking.

It seems pssh get some retries implemented by default: https://parallel-ssh.readthedocs.io/en/0.80.0/pssh_client.html

```
classpssh.pssh_client.ParallelSSHClient(hosts, user=None, password=None, port=None, pkey=None, forward_ssh_agent=True, num_retries=3, timeout=120, pool_size=10, proxy_host=None, proxy_port=22)
```

This might be enough. If not we can always add some retry logic here: https://github.com/thelastpickle/cassandra-medusa/pull/228/files#diff-f3de2f83d580df09f44abd0588b87a07d5a35e2bd462d46d69b4a983dd333e42R158-R162

I think we can try to rely on pssh retries as a first intention and implement a custom retry logic only if you (or we) see problems when commands fail.

adejanovski (approved):
LGTM aside from a minor nit

medusa/cassandra_utils.py
```diff
@@ -364,6 +363,8 @@ def rpc_port(self):
         return self._rpc_port
 
     class Snapshot(object):
+        prefix = 'medusa-'
```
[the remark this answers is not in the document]
arodrime: Fair enough.

PR #232, opened 2020-11-25
Build ubuntu package for focal 20.04

masterzen:
This adds the build configuration for the ubuntu 20.04 focal debian package, along with the other debian packages built.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,4 @@
 debian/cassandra-medusa/
+packages/
+.direnv/
+.venv/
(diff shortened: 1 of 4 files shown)


adejanovski:
Hi @masterzen,

great addition, thanks!
I'm thinking that maybe we could add a CI step to build the packages even when we're not doing a release, and check that we get the expected packages as output.
That would help secure this process and avoid errors popping up when we try to release Medusa 😅 

What do you think?

adejanovski:
Merged as part of #243

PR #237, opened 2020-12-14
Add ability to use C* Management API for snapshot management

emerkle826:
This PR adds the ability to use the Cassandra [Management API]() in a Kubernetes environment for performing backup and restore functionality, instead of using JMX and needing to have a Jolokia agent embedded into your Cassandra image. This capability is intended for a K8s setup, and requires specific configuration in Medusa to enable.

Included in this PR is a slight refactoring of the `create snapshot` and `delete snapshot` functions, as well as introducing a _snapshot service_ abstraction layer to allow for multiple implementations to manage snapshots in Cassandra.

This should address #262

### DIFF
--- a/tests/integration/features/environment.py
+++ b/tests/integration/features/environment.py
@@ -40,3 +40,8 @@ def after_step(context, step):
         # NOTE: Use IPython debugger, same for pdb (basic python debugger).
         import ipdb
         ipdb.post_mortem(step.exc_traceback)
+
+
+def before_scenario(context, scenario):
+    if "skip-cassandra-2" in scenario.effective_tags and context.cassandra_version.startswith("2."):
+        scenario.skip("Skipping scenario on Cassandra 2.x")
--- a/tests/integration/features/integration_tests.feature
+++ b/tests/integration/features/integration_tests.feature
@@ -686,7 +686,7 @@ Feature: Integration tests
         | minio             | without_client_encryption |
 
     @16
-    Scenario Outline: Perform a differential backup over gRPC , verify its index, then delete it over gRPC
+    Scenario Outline: Perform a differential backup over gRPC , verify its index, then delete it over gRPC with Jolokia
         Given I have a fresh ccm cluster with jolokia "<client encryption>" running named "scenario16"
         Given I am using "<storage>" as storage provider in ccm cluster "<client encryption>" with gRPC server
         Then the gRPC server is up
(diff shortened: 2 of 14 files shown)


tests/integration/features/integration_tests.feature
```diff
@@ -649,3 +649,26 @@ Feature: Integration tests
         Examples: Local storage
         | storage           | client encryption |
         | local      |  with_client_encryption |
+
+    @17 @skip-cassandra-2
```
emerkle826, line 653: The reason we skip this test for Cassandra 2.x is that Management API is only supported on Cassandra 3.x and up.

tests/integration/features/steps/integration_steps.py
```diff
@@ -1114,3 +1390,21 @@ def write_dummy_file(path, mtime_str, contents=None):
...
+        for chunk in zip_file.iter_content(chunk_size=4096):
+            if chunk:
+                mgmt_api_jars.write(chunk)
+
+    with ZipFile('/tmp/mgmt_api_jars.zip', 'r') as zip_ref:
+        list_of_names = zip_ref.namelist()
+        for file_name in list_of_names:
+            if file_name.endswith('.jar'):
+                if 'mgmtapi-agent' in file_name or 'mgmtapi-server' in file_name:
+                    zip_ref.extract(file_name, '/tmp')
```
emerkle826, line 1410: This is just a test utility method to pull down the Management API jars so they can be unzipped and used to embed into the running CCM Cassandra instance for the test. It might be worth it to configure the version to use (it's called on line 334 [above](https://github.com/thelastpickle/cassandra-medusa/pull/237/files#diff-e513e4ce15a41a09a27c952d7e5d7afc4a04ed85d3022707ca912ab26c5cce14R334)). For now I have it defaulted to v0.1.17 here (the current latest).

tests/integration/features/environment.py
```diff
@@ -40,3 +40,8 @@ def after_step(context, step):
         # NOTE: Use IPython debugger, same for pdb (basic python debugger).
         import ipdb
         ipdb.post_mortem(step.exc_traceback)
+
+
+def before_scenario(context, scenario):
+    if "skip-cassandra-2" in scenario.effective_tags and context.cassandra_version.startswith("2."):
```
adejanovski, line 46: Cool trick 🙂

adejanovski (dismissed):
@emerkle826, this is a fantastic PR 😍 
The refactoring of the snapshot service is beautifully designed and separates concerns in a very clean way.

Thanks and approved (finally)!

adejanovski (approved):

Issue #236, opened 2020-12-11
Using offline s3 storage such as minio


DrissiReda:
Is it possible to use another s3 storage than classic aws? I need to upload my backups remotely to an offline remote storage, I've seen in the code that the [storage] section supports the option "host" but it seems that the storage_provider s3 just ignores it.

Can anyone point me into the right direction as to how I can achieve this?

adejanovski:
Hi @DrissiReda,

I don't think you can achieve this through configuration only.
You may want to take a look at [this PR](https://github.com/thelastpickle/cassandra-medusa/pull/226) which added support for IBM Cloud Storage to guide your implementation. IBM Cloud Storage is compatible with S3, including with the awscli but it requires to add custom endpoint urls depending on the region.
You'll have to create a new libcloud provider that will inherit from the S3 provider and then create a MinioStorage class under `medusa/storage`.

DrissiReda:
Thanks @adejanovski I made a PR for it.

jsanda:
Hi @DrissiReda,

It is by design that nodetool is not in image used to run the gRPC service. When Medusa is run inside Kubernetes, it won't have access to nodetool. It makes REST api requests to the management-api (used by cass-operator) or to Jolokia for creating snapshots.

The image is really only intended for use inside of Kubernetes. I would not expect it to work outside of a Kubernetes deployment.

burmanm:
I think it would make more sense to just allow S3 to use the host/port or endpoint_url. And then merge the IBM Cloud to that one and use the same for Minio. All those S3 compatible storages should work that way.

DrissiReda:
You would need to make the libcloud driver for it as well.

burmanm:
Why? Just set the driver to be s3 in the config.

Skunnyk:
We have a custom s3 infra/endpoint, so it would be marvelous to be able to connect to it simply by overriding some variables.
With `awscli` for example, you can override the endpoint with the https://pypi.org/project/awscli-plugin-endpoint/ plugin.

Reimplementing a "driver" for every compatible s3 provider is not really "kiss". But maybe it's tied to how `libcloud` works…

adejanovski:
There are two sides to this:
- Libcloud in its current design won't allow us to support a new S3 compatible backend without creating a specific subclass of the S3 driver.
[...]

PR #246, opened 2021-01-04
Added minio support

DrissiReda:
Solves #236


burmanm:
I don't see the point in having ``ibm_cloud_storage/`` and ``minio_storage/`` directories with nearly similar files. I think all these S3-compatible storages should have a common directory and only override something if necessary. For example, these two directories have nearly no difference (one does self.endpoint_url formatting while other one does it later - but they could both use the self) other than the selection of region pointing to URL. 

Having less copies of the same functionality makes it easier to maintain these. There's plenty of S3 compatible clones and I guess we'd like to support all of them through some ``s3compat`` type of driver. That should be somewhat easy as we only use few features so the difference between say Minio and DigitalOcean's Spaces, B2 Backblaze, etc shouldn't make any difference to us. I'd say we probably don't even need to update the libcloud and use the minio driver from there, but perhaps just making the ibm_cloud_storage code to be generic s3compat would do the job (excluding the region autofilling). To me setting the endpoint url should be fine override to provide lowest possible work to connect to any provider. @adejanovski would you oppose such "quick fix" ? It would make testing easier as the docker compose or equal could spin up minio for testing the backups against s3 / awscli.

DrissiReda:
Indeed if we could expose the endpoint url and the region in the configuration, it could work. Make the region be autofilled if the setting wasn't provided. Hardcoding the region will definitely break some otherwise compatible s3 backends.

david316:
I thought that it might be relevant to let you know that I've successfully used the Ceph s3_rgw provider with Minio. 

I agree with the concept of a more generic s3 compatible provider that accepts an endpoint and region configuration.

adejanovski:
> To me setting the endpoint url should be fine override to provide lowest possible work to connect to any provider. @adejanovski would you oppose such "quick fix" ? It would make testing easier as the docker compose or equal could spin up minio for testing the backups against s3 / awscli.

@burmanm, I'm happy with your suggestion. We already have a `host` setting that is used for the endpoint in the s3_rgw implementation and the region is available already as well. The IBM storage code seems fit as base for an `S3Generic`/`S3Compatible` class.

We will then also merge the `concurrent.py` and `awscli.py` variants into a single one by using the `ibm_cloud_storage` one and adapting it for the AWS S3 case (ie removing the `endpoint-url` and `region` arguments in the aws cli calls).

adejanovski:
@DrissiReda, the newly merged support for S3 compatible implementations should provide Minio access.
Would you mind testing `master` with Minio and tell us if it works ok for you?
Seems like we've been unable to reproduce the region issue you've seen.
Thanks!

adejanovski:
Closing this PR as MinIO support was merged to `master` through another PR.
Thanks a lot @DrissiReda for pushing this PR nonetheless!

PR #255, opened 2021-01-13
Added credential insertion

DrissiReda:
For kubernetes case we need to be able to read cassandra credentials from environment for multiple reasons:
- when user creation is automated, we cannot always know the credentials prior to deployment.
- having secrets is always better than having plain text.
This change in the docker entrypoint will read from environment variables `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`, then either add them to `/etc/medusa/medusa.ini` if they don't exist, or replace them if they do.

**NOTE**: if you only supply a username, the password will just be an empty string, even if the `/etc/medusa/medusa.ini` already had a password, it'll be overwritten with an empty one.

### DIFF
--- a/k8s/docker-entrypoint.sh
+++ b/k8s/docker-entrypoint.sh
@@ -50,7 +50,25 @@ grpc() {
     echo "Starting Medusa gRPC service"
     python3 -m medusa.service.grpc.server server.py
 }
-
+# Add username/password to the medusa-config
+if [ ! -z $CASSANDRA_USERNAME ]; then
+    if ! grep "^\[cassandra\]" /etc/medusa/medusa.ini > /dev/null ; then
+        echo "[cassandra]" >> /etc/medusa/medusa.ini
+    fi
+    if grep "^cql_username" /etc/medusa/medusa.ini > /dev/null; then
+        sed -i "s/^cql_username.*/cql_username = $CASSANDRA_USERNAME/g" /etc/medusa/medusa.ini
+    else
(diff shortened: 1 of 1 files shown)


jsanda:
Hi @DrissiReda 

I read through https://github.com/kubernetes/kubernetes/issues/79224 which is very relevant to this situation. Unfortunately there is not a good way to add secrets to a ConfigMap. Based on what I read in that ticket I think my preference would be to add support in Medusa for getting the credentials from env vars. Maybe we could add support for a couple properties in `medusa.ini`, something like `cql_username_env_var` and `cql_password_env_var`. Values should specify the names of the environment variables that store the credentials.

@adejanovski what do you think about this idea?

DrissiReda:
Thanks for the review john, I know it's not ideal, and I've seen why kubernetes prevents configMaps from being writable. But without touching medusa's code, I couldn't think of a better solution.

jsanda:
> Thanks for the review john, I know it's not ideal, and I've seen why kubernetes prevents configMaps from being writable. But without touching medusa's code, I couldn't think of a better solution.

I hear ya :) Medusa was not built with Kubernetes in mind. There may be situations where it makes sense to introduce changes in Medusa to better support running in Kubernetes. I think this is one of those situations.

adejanovski:
@DrissiReda, we merged #263, which makes Medusa use the `CQL_USERNAME` and `CQL_PASSWORD` env variable if they exist.
Would that work for you in replacement of this PR?

DrissiReda:
Oh yes definitely, thanks !

PR #259, opened 2021-01-19
Unthrottle downloads for S3 and IBM storage backends to speed up restores

adejanovski:


medusa/storage/ibm_storage.py
```diff
@@ -118,16 +118,3 @@ def download_blobs(self, srcs, dest):
 
 def is_aws_s3(storage_name):
     return True
-
-
-def prepare_download(self):
```
rzvoncek, line 123: So if the `ibm_storage` doesn't have this, it'll use the `abstract_storage`'s method? Right? We wont be getting any exceptions or anything.

rzvoncek (approved):

medusa/storage/ibm_storage.py
adejanovski: Yes, I'll run the integration tests using the IBM backend to make sure of this.

adejanovski: It's passing nicely: https://github.com/thelastpickle/cassandra-medusa/runs/1785219807?check_suite_focus=true

I'll drop that last commit and merge the PR.

PR #263, opened 2021-01-22
Allow to set the CQL user/pass through env variables

adejanovski:


medusa/cassandra_utils.py
```diff
@@ -59,7 +59,8 @@ def __init__(self, ip_addresses, cassandra_config):
         self._ssl_context = None
         self._cassandra_config = cassandra_config
 
-        if cassandra_config.cql_username is not None and cassandra_config.cql_password is not None:
+        if null_if_empty(cassandra_config.cql_username) is not None \
```
jsanda, line 62: Could these checks be simplified with:

```python
if cassandra_config.cql_username and cassandra_config.cql_username:
  ...
```

jsanda (approved):
I left one question, but this looks good to me. These changes will make https://github.com/k8ssandra/k8ssandra/issues/154 a lot easier. Thanks!

medusa/cassandra_utils.py
adejanovski: You got it 👌

PR #265, opened 2021-01-27
Backup and DeleteBackup do not return error code

burmanm:
In case of Exception, Backup and DeleteBackup methods in the gRPC should return an error code. Fixes https://github.com/k8ssandra/k8ssandra/issues/256

### DIFF
--- a/medusa/service/grpc/server.py
+++ b/medusa/service/grpc/server.py
@@ -46,14 +46,16 @@ class MedusaService(medusa_pb2_grpc.MedusaServicer):
 
     def Backup(self, request, context):
         logging.info("Performing backup {}".format(request.name))
+        resp = medusa_pb2.BackupResponse()
         # TODO pass the staggered and mode args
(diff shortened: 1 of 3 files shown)


adejanovski (changes requested):
LGTM.

Would you mind adding a test for this? 
Scenario 16 of the integration tests could be reused as a base to generate a failing backup and test the return code.
Thanks!

jsanda:
I will manually test this in k8ssandra and report back my results.

jsanda:
The changes look good. I reported my results in https://github.com/k8ssandra/k8ssandra/issues/256.

@burmanm after you add the integration test and this PR is merged, we will need to update the Medusa image in k8ssandra before we can close out https://github.com/k8ssandra/k8ssandra/issues/256.

adejanovski (approved):
Thanks for adding the tests @burmanm !

LGTM

Issue #184, opened 2020-07-30
Need to have --region and --debug parameter for aws s3 cp


chidhambaram7:
Hi,

We have decided to use AWS S3 as backup location for our clusters. So, we use AWS VPC S3 endpoint to upload and download data from S3. To use VPC endpoint, I have to sepcify `--region` parameter along with `aws s3 cp`. As of now, I have modified `awscli.py` file manually with the required parameter. 
Please consider to have more parameter to append to `aws s3 cp` command.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-292) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-292

sigalits:
I have also came across that issue. will be happy to get it resolved

adejanovski:
The `--region` param was added in #272

PR #272, opened 2021-02-02
Consolidate S3 compatible storage options, add s3c target

burmanm:
Also, AWS s3 driver can accept region parameter (https://github.com/k8ssandra/k8ssandra/issues/142). The new name for the driver was ``s3c`` although we can change this. I picked the MinIOStorageDriver as the base driver, since it seems to be the native S3 driver simply extended with init: https://github.com/apache/libcloud/blob/trunk/libcloud/storage/drivers/minio.py

Keeps the ``s3_rgw``, if someone wants to build a Docker-image without the awscli dependency (to reduce storage size) or otherwise wishes to use the libcloud's version of s3 driver. 

Fixes https://github.com/thelastpickle/cassandra-medusa/issues/236
Fixes https://github.com/thelastpickle/cassandra-medusa/issues/184 (not the --debug part though)

Still needs testing before merging.


medusa/storage/s3_compat_storage/concurrent.py
```diff
@@ -24,7 +24,7 @@
 from retrying import retry
 
 import medusa
-from medusa.storage.aws_s3_storage.awscli import AwsCli
+from medusa.storage.s3_compat.awscli import AwsCli
```
adejanovski, line 27:
```suggestion
from medusa.storage.s3_compat_storage.awscli import AwsCli
```

medusa/storage/s3_compat.py
```diff
@@ -0,0 +1,202 @@
+# -*- coding: utf-8 -*-
```
adejanovski, line 1: Could you rename this file to `s3_base_storage` instead? While it's not the case everywhere, I think it would be good to standardize file names with their main class name from now on like we would in Java.
Also `base` sounds more accurate since it also serves as the base for the "real" AWS S3 implementation.

medusa/storage/s3_compat.py
```diff
@@ -0,0 +1,202 @@
...
+# Copyright 2019 Spotify AB
```
adejanovski, line 2:
```suggestion
# Copyright 2019 Spotify AB
# Copyright 2021 DataStax, Inc.
```

adejanovski (changes requested):
Looks great so far!
And it feels good to remove a lot of duplicated code, thanks for that 🙏 

I have a few minor comments and a little concern with the Libcloud MinIO driver with region shenanigans. We'd need to clear this out with some testing maybe?

medusa/storage/s3_compat.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class S3BaseStorage(AbstractStorage):
+
+    def connect_storage(self):
+        with io.open(os.path.expanduser(self.config.key_file), 'r', encoding='utf-8') as json_fi:
+            credentials = json.load(json_fi)
+
+        # MinIOStorageDriver is the only clean implementation of BaseS3StorageDriver in libcloud
+        driver = MinIOStorageDriver(
```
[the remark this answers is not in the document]
DrissiReda: > 
> 
> Apparently there was [an issue](https://github.com/apache/libcloud/issues/1539) with the region setting in the MinIOStorageDriver, which was fixed by creating [this custom class](https://github.com/thelastpickle/cassandra-medusa/pull/246/files#diff-c3d3aefd89ed037caf0585e9927e2628eab86f6e33e2ac61fb77683bd9d748ce).
> What's your take on this? Shouldn't we use the driver implemented in #246 for now, until this is clarified?
> @DrissiReda, I saw you closed [the libcloud issue](https://github.com/apache/libcloud/issues/1539). Is it because it was related to a configuration problem or was it because you were happy with the custom driver you created for MinIO in your medusa PR?

Custom driver wasn't enough, like I said in that issue, the libcloud driver overwrites the region as an empty string, so I had to modify that for it to work, even in libcloud it's not exposed to the configuration, so a hardcoded fix like what I did is needed for minIO, for other s3 providers, you can't do this, since the region can change, and you'll have to expose it as a parameter. 

This is the fix I'm referring to:
 Setting https://github.com/apache/libcloud/blob/4370a7021c11c1c61afe6470d2baafc093bfa0dd/libcloud/storage/drivers/minio.py#L49 
`region_name = "us-east-1"`

burmanm:
It seems I pushed the commit from https://github.com/thelastpickle/cassandra-medusa/pull/272 also (with some rebase fixes), but I tried to verify it works also:

```
⋊> ~/p/g/d/a/dist ./aws s3 --endpoint-url=http://192.168.1.201:9000 ls s3://awstest/                                                                                           18:51:52
                           PRE index/
                           PRE k8ssandra-dc1-default-sts-0.k8ssandra-dc1-service.default.svc.cluster.local/
⋊> ~/p/g/d/a/dist   
```

medusa/storage/s3_compat.py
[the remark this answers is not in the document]
burmanm: I don't think we use the driver in cases where that happens, since the awscli is used for download / upload?

medusa/storage/s3_compat.py
burmanm: Sure

medusa/storage/s3_compat.py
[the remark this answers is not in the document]
burmanm: At least I didn't hit it yet..

jsanda:
@adejanovski after @burmanm merges/rebases can you please do another review on this? This PR includes a fix for https://github.com/k8ssandra/k8ssandra/issues/142 which we need to get done for k8ssandra 1.0.

PR #267, opened 2021-01-28
Azure Blobs fixes -  fix differential backups & specify host parameter

mkcello96:
2 fixes:

1) Fix differential backups - I was getting an error that `hash_in_manifest` was None on [this line](https://github.com/thelastpickle/cassandra-medusa/pull/219/files#diff-1eca9b6310a202f9d27714f7f0330d78c90edd3445ed2ed0c3872aa165075e7dR97) when trying to run a differential backup with large files.
The fix is similar to https://github.com/thelastpickle/cassandra-medusa/pull/27, when differential backups were broken in s3. I simply copied the multipart file code currently used by s3 to fix the issue.

2) Our storage accounts are in Azure Gov so I added support to specify a `host` in the libcloud AzureBlobsStorageDriver connection, as well as a `connection_string` (which contains host information) in the AWS-cli connection. Unfortunately libclould has a bug in this scenario, so I used a workaround that I found and linked the issue I filed with them in the code (https://github.com/apache/libcloud/issues/1551) 

I verified the fix works by running a differential backup to Azure w/ multi-part files, which resulted in skipping the already-uploaded files. The integration tests should be updated as well - to test the new `host` connections you can simply set them in the `medusa-azure-credentials` file and run the azure tests.

PR #278, opened 2021-02-06
Azure Blobs fixes - fix differential backups & specify host parameter

mkcello96:
(same PR as https://github.com/thelastpickle/cassandra-medusa/pull/267 - I made this one to clean the git history, as I had merged master a few times). 

2 fixes:

Fix differential backups - I was getting an error that hash_in_manifest was None on this line when trying to run a differential backup with large files.
The fix is similar to #27, when differential backups were broken in s3. I simply copied the multipart file code currently used by s3 to fix the issue.

Our storage accounts are in Azure Gov so I added support to specify a host in the libcloud AzureBlobsStorageDriver connection, as well as a connection_string (which contains host information) in the AWS-cli connection. Unfortunately libclould has a bug in this scenario, so I used a workaround that I found and linked the issue I filed with them in the code (apache/libcloud#1551)

I verified the fix works by running a differential backup to Azure w/ multi-part files, which resulted in skipping the already-uploaded files. The integration tests should be updated as well - to test the new host connections you can simply set them in the medusa-azure-credentials file and run the azure tests.

### DIFF
--- a/docs/azure_blobs_setup.md
+++ b/docs/azure_blobs_setup.md
@@ -11,6 +11,14 @@ Create a new storage account or use an existing one which will be used to store 
     "key": "YOUR_KEY"
 }
 ```
+If you need to set a different host for Azure (for example the host for Azure Gov is `<storageAccount>.blob.core.usgovcloudapi.net`), please ADDITIONALLY set these two fields in the JSON file (the connection string can be found with the key):
+
+```
+"host": "YOUR_HOST"
+"connection_string": "YOUR_CONNECTION_STRING"
+
+```
+
 Place this file on all Cassandra nodes running medusa under `/etc/medusa/`and set the rigths appropriately so that onyl users running Medusa can read/modify it.
 
 ### Create a container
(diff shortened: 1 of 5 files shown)


mkcello96:
@adejanovski Thanks for pointing out the build failure, I fixed that and added one more thing to this PR - the MD5 hash needs to be specified for azcli blob uploads (azcli handles multi-part uploads). This is currently breaking the `verify` command for Azure backups with large files (since it expects the MD5). I tested this fix by running `verify` on a full Azure backup. 

(this is the same PR as #267 , I just cleaned up the git history after I learned not to merge master) :)

adejanovski (dismissed):
LGTM @mkcello96 and I successfully ran the whole integration test suite using the Azure backend.

Could you squash your 7 commits into a single one before I merge the PR?

To do this, run: `git rebase -i HEAD~7`
then replace `pick`  with `squash` for all commits but the first one, then save with `:wq`.
Rework the commit messages and finally force push the branch with `git push --force-with-lease`.

Thanks!

mkcello96:
@adejanovski Done, thanks for your help on this one!

adejanovski (approved):

mkcello96:
@adejanovski FYI, the issue I opened with libcloud (apache/libcloud#1551, to fix the host connection to Azure) has a PR to resolve the issue. Unfortunately it's not compatible with my workaround that was merged in this PR. 
I'll plan on circling back around to this when that libcloud fix gets released. Hoping we can deprecate older libcloud versions when that happens.

adejanovski:
Hi @mkcello96,
FTR, we've limited the Libcloud versions to be between 2.8.0 and 3.3.0, so we're protected from any release that would break Medusa.
Glad to hear you're on top of things though and let me know if you need help with anything.

PR #283, opened 2021-02-09
Add s3_compatible storage option for all S3 compatible storage targets

burmanm:
Further consolidation of the code, s3_compatible is the new s3 compatible storage driver

Fix some bugs, typos and make the s3_compat load credentials with the aws keys (for compatibility)

Address comments, add http/https differentation to endpoint_url generation

Remove is_azure from Azure blobs storage as backup_node does not need it
(if threshold is set, it will be used)

(cherry picked from commit 544940dda0463df7e062113ad498345fa97a5955)


burmanm:
This is the version for the CI runs.

medusa/storage/s3_storage.py
```diff
@@ -12,29 +12,19 @@
...
+from medusa.storage.s3_base_storage import S3BaseStorage
 
-import medusa
 
-
-class S3Storage(AbstractStorage):
+class S3Storage(S3BaseStorage):
     """
     Available storage providers for S3:
     S3_AP_NORTHEAST = 's3_ap_northeast'
```
adejanovski, line 30: We should remove these in favor of using `s3` as storage provider and setting the region instead of using region specific storage providers.

medusa/storage/ibm_storage.py
```diff
@@ -59,62 +54,7 @@ def connect_storage(self):
...
-        try:
-            subprocess.check_call([aws_cli_path, "help"], stdout=PIPE, stderr=PIPE)
-        except Exception:
-            raise RuntimeError(
-                "AWS cli doesn't seem to be installed on this system and is a "
-                + "required dependency for the S3 backend. Please install it by running 'pip install awscli' "
-                + "or 'sudo apt-get install awscli' and try again."
-            )
+        if self.config.host is None and self.config.region is not None:
+            self.config.host = IBM_CLOUD_HOSTS_BY_REGION[self.config.region]
```
adejanovski, line 58: You cannot change a config element like this. But I think we should drop this static mapping and let the user fill in the host and region in config.

adejanovski (changes requested):
The IBM storage tests are currently failing due to how you try to override `config.host`. 
I think we should drop the specific bits for the IBM storage backend and fully rely on the new S3 generic class, removing the host/region IBM mapping which uses a hardcoded list anyway (that's on me).
Let's still identify `ibm` as a valid storage provider and use the `S3BaseStorage` class for it directly like it's done for `s3_compatible`.

adejanovski (approved):

PR #285, opened 2021-02-17
Fix home/medusa directory with correct rights

burmanm:
Also remove unnecessary app directory


jsanda (approved):
Tested and verified changes in k8ssandra 👍

PR #302, opened 2021-03-04
Add BaseS3StorageDriver that allows defining the region

burmanm:
Also pass the information to that driver. And remove an unused IBM driver. I couldn't get the integration tests to work on my machine, so I have to rely on CircleCI now..


adejanovski (changes requested):
@burmanm, PR looks great, thanks!
Would it be possible to rebase on top of 0.9 and point the PR to that branch instead of `master`? 
I think this PR matches the definition of a bug fix.

@Skunnyk, would you be able test this PR and confirm it works correctly with OVHcloud S3?
That would be super helpful 🙏

adejanovski:
@burmanm, once the rebase is done, could you squash the commits so that we're ready to rebase and merge? Thanks

Issue #303, opened 2021-03-04
Add support for setting tokens during restore when running in Kubernetes


jsanda:
In k8ssandra restoring to a new cluster currently does not work (see https://github.com/k8ssandra/k8ssandra/issues/70). There are a couple issues that need to be resolved. One of those issues is that Medusa does not update tokens in cassandra.yaml when running in Kubernetes. Here is the relevant code from the `restore_node_locally` function in `medusa/restore_node.py`:

```python
    if not medusa.utils.evaluate_boolean(config.kubernetes.enabled):
        if seeds is not None:
            wait_for_seeds(config, seeds)
        else:
            logging.info('No --seeds specified so we will not wait for any')

        # Start up Cassandra
        logging.info('Starting Cassandra')
        # restoring in place retains system.local, which has tokens in it. no need to specify extra
        if in_place:
            cassandra.start_with_implicit_token()
        else:
            cassandra.start(tokens)
```
The code that updates cassandra.yaml happens in `cassandra.start(tokens)`.  We need to be able to update cassandra.yaml without starting Cassandra (since Medusa won't be able to start Cassandra in Kubernetes).

adejanovski:
A couple of other things that need to be done for a remote restore:

- Disable auto bootstrap
- Prevent restoring the `system.local` table (Haven't checked the code for k8s specific paths so maybe it's already handled).

jsanda:
Thanks for the info @adejanovski.

There does not appear to be any k8s specific paths around `system.local`, so that's good.

burmanm:
Does that system.local need any change?

There's a code that system.local is never restored for not in place: https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/restore_node.py#L283

adejanovski:
No change needed indeed, Medusa already deals with this correctly, including the k8s case.

PR #308, opened 2021-03-16
If running in Kubernetes, replace tokens in not in_place case

burmanm:
And also disable bootstrap (call the same command as start would usually do, but do not start the node).

Related: #303

### DIFF
--- a/medusa/restore_node.py
+++ b/medusa/restore_node.py
@@ -132,6 +132,9 @@ def restore_node_locally(config, temp_dir, backup_name, in_place, keep_auth, see
             cassandra.start_with_implicit_token()
         else:
             cassandra.start(tokens)
+    elif not in_place:
+        # Kubernetes will manage the lifecycle, but we still need to modify the tokens
+        cassandra.replaceTokensInCassandraYamlAndDisableBootstrap(tokens)
 
     # Clean the restored data from local temporary folder
     clean_path(download_dir, use_sudo, keep_folder=False)


medusa/restore_node.py
```diff
@@ -131,7 +131,10 @@ def restore_node_locally(config, temp_dir, backup_name, in_place, keep_auth, see
         if in_place:
             cassandra.start_with_implicit_token()
         else:
-            cassandra.start(tokens)
+            if medusa.utils.evaluate_boolean(config.kubernetes.enabled):
```
adejanovski, line 134: This block is already bypassed in the case of k8s due [to this condition](https://github.com/thelastpickle/cassandra-medusa/pull/308/files#diff-558c6e6db1fe2499e845c25c2379d77a10d0529846a0ee9d3e54931cd9b25d2dR122).

I think we rather need to add an else clause: 

```
    if not medusa.utils.evaluate_boolean(config.kubernetes.enabled):
        if seeds is not None:
        ...
        ...
    elif not in_place:
      # kubernetes remote restore
      cassandra.replaceTokensInCassandraYamlAndDisableBootstrap(tokens)
```

What do you think?

adejanovski (changes requested):
A small change is required to get this working. 
Happy to see remote restore for k8s making progress 😊

medusa/restore_node.py
burmanm: Ouh, my blind eyes.

adejanovski (approved):
LGTM now 👍

adejanovski:
@burmanm, could you please squash/fixup before merging?

PR #309, opened 2021-03-16
Rebrand to Medusa for Apache Cassandra

emerkle826:
I also have changes to the Wiki, though a GH PR isn't easy to create for Wiki page updates. You can see my changes here:
https://github.com/emerkle826/cassandra-medusa-wiki/commit/9af7ddff00b34ebf4d6e2c9cc9b93696150fc2e4

~NOTE: The images in the Wiki likely need to be updated to say "Medusa for Apache Cassandra" as well, though I have not made those edits yet.~
Images have been updated with the latest commit.


adejanovski (changes requested):
Thanks for making all these changes @emerkle826 !
I have a few comments about the CoC and the fact that Medusa for Apache Cassandra should mostly replace mentions of "Cassandra Medusa" which is a copyright infringement.
Also, the rule is to have the first mention of Cassandra in a document be "Apache Cassandra™" and then use just Cassandra for the subsequent mentions in that same document.

emerkle826:
@adejanovski I've reworked this a bit following your suggestions. It turns out that there really aren't any references to "Cassandra Medusa" anywhere in the docs. I've also reverted the Code of Conduct for now, so it still mentions Spotify OSS. I think that can be addressed if/when the project is moved.

emerkle826:
Closing this in favor of adding the changes to the branch here:
https://github.com/thelastpickle/cassandra-medusa/tree/alex/docs-in-tree

PR #310, opened 2021-03-19
Add new parameter to allow overriding the fqdn for new cluster

burmanm:
If restoring to a new cluster, we need to fetch the backup from an old name (instead of the new cluster name / podname).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1407) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1407
┆priority: Medium

### DIFF
--- a/medusa/restore_node.py
+++ b/medusa/restore_node.py
@@ -37,7 +37,7 @@ MAX_ATTEMPTS = 60
 
 
 def restore_node(config, temp_dir, backup_name, in_place, keep_auth, seeds, verify, keyspaces, tables,
-                 use_sstableloader=False):
+                 use_sstableloader=False, backup_fqdn=None):
 
     if in_place and keep_auth:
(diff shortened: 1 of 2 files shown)


burmanm:
15 lines max per function in SonarCloud settings? And why does it complain duplication from the project for PRs that do not modify the files?

adejanovski:
> 15 lines max per function in SonarCloud settings? 

It's not complaining about the number of lines actually :) Sonar computes a cognitive complexity for methods/functions, and has a default threshold at 15. If we're over this, it means there are too many things happening in the same function, making it hard to understand. It should then be broken up into smaller functions to help making it readable.

> And why does it complain duplication from the project for PRs that do not modify the files?

You're right, it shouldn't be complaining here so let's ignore this. You clearly didn't touch any of the files that are reported.

PR #313, opened 2021-03-21
add handler for sigterm and use exec to run server.py script

jsanda:
This PR includes a couple changes:

* Adds a signal handler for SIGTERM
    * When invoked the handler logs that the server is shutting down
    * The handler calls `server.stop()` on the gRPC serve
* Use `exec` to run `server.py` so that it is not run in a subshell

### DIFF
--- a/k8s/docker-entrypoint.sh
+++ b/k8s/docker-entrypoint.sh
@@ -48,7 +48,7 @@ restore() {
 
 grpc() {
     echo "Starting Medusa gRPC service"
-    python3 -m medusa.service.grpc.server server.py
+    exec python3 -m medusa.service.grpc.server server.py
 }
 
 echo "sleeping for $DEBUG_SLEEP sec"
@@ -57,7 +57,8 @@ sleep $DEBUG_SLEEP
 if [ "$MEDUSA_MODE" == "RESTORE" ]; then
     restore
 elif [ "$MEDUSA_MODE" == "GRPC" ]; then
-    grpc
+    echo "Starting Medusa gRPC service"
+    exec python3 -m medusa.service.grpc.server server.py
 else
     echo "MEDUSA_MODE env var must be set to either RESTORE or GRPC"
     exit 1
--- a/k8s/docker-entrypoint.sh
+++ b/k8s/docker-entrypoint.sh
@@ -41,7 +41,7 @@ restore() {
         echo "Skipping restore operation"    
     else
         echo "Restoring backup $BACKUP_NAME"
-        python3 -m medusa.service.grpc.restore
+        exec python3 -m medusa.service.grpc.restore
         echo $RESTORE_KEY > $last_restore_file
     fi
 }
@@ -57,8 +57,7 @@ sleep $DEBUG_SLEEP
 if [ "$MEDUSA_MODE" == "RESTORE" ]; then
     restore
 elif [ "$MEDUSA_MODE" == "GRPC" ]; then
-    echo "Starting Medusa gRPC service"
-    exec python3 -m medusa.service.grpc.server server.py
+    grpc
 else
     echo "MEDUSA_MODE env var must be set to either RESTORE or GRPC"
     exit 1
--- a/medusa/service/grpc/server.py
+++ b/medusa/service/grpc/server.py
@@ -17,6 +17,7 @@ import logging
 import os
 import sys
 import time
+import signal
 from collections import defaultdict
 from concurrent import futures
 from datetime import datetime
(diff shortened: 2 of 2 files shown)


medusa/service/grpc/server.py
```diff
@@ -162,10 +167,14 @@ def configure_console_logging(config):
...
+server.wait_for_termination()
+
 # since server.start() will not block,
 # a sleep-loop is added to keep alive
-try:
-    while True:
-        time.sleep(86400)
-except KeyboardInterrupt:
-    server.stop(0)
+# try:
```
adejanovski, line 176: Would it be possible to remove these lines if they're not used anymore?

k8s/docker-entrypoint.sh
```diff
@@ -48,7 +48,7 @@ restore() {
 
 grpc() {
     echo "Starting Medusa gRPC service"
-    python3 -m medusa.service.grpc.server server.py
+    exec python3 -m medusa.service.grpc.server server.py
```
adejanovski, line 51: Out of curiosity, what's the difference with using `exec` instead of calling `python3` directly? Is it related to signal handling?

adejanovski (changes requested):
There's a comment that looks like it could be removed and maybe it could be nice to have a small test somewhere that starts the server and then sends a SIGTERM to verify that the change works while protecting us from future regression.
wdyt?

medusa/service/grpc/server.py
jsanda: >There's a comment that looks like it could be removed and maybe it could be nice to have a small test somewhere that starts the server and then sends a SIGTERM to verify that the change works while protecting us from future regression.
wdyt?

Definitely. I will work on a test.

>Would it be possible to remove these lines if they're not used anymore?

👍

k8s/docker-entrypoint.sh
jsanda: `exec` causes the new process to be run with in the current shell/process.

Here is the `ps` output from within a `medusa` container that does **not** use `exec`:

```
$ ps -x
    PID TTY      STAT   TIME COMMAND
      1 ?        Ss     0:00 /bin/bash /home/cassandra/docker-entrypoint.sh
      7 ?        Sl     0:00 python3 -m medusa.service.grpc.server server.py
    372 pts/0    Ss     0:00 /bin/bash
```
Running `kill -SIGTERM 1` has no effect here. 

Here is the `ps` output from a `medusa` container where `exec` is used:

```
$ ps -x
    PID TTY      STAT   TIME COMMAND
      1 ?        Ssl    0:00 python3 -m medusa.service.grpc.server server.py
    347 pts/0    Ss     0:00 /bin/bash
```
Running `kill -SIGTERM 1` has the desired effect here.

k8s/docker-entrypoint.sh
```diff
@@ -57,7 +57,8 @@ sleep $DEBUG_SLEEP
 if [ "$MEDUSA_MODE" == "RESTORE" ]; then
     restore
 elif [ "$MEDUSA_MODE" == "GRPC" ]; then
-    grpc
+    echo "Starting Medusa gRPC service"
+    exec python3 -m medusa.service.grpc.server server.py
```
emerkle826, line 61: If changing the `grpc()` function above in this script has the desired affect, can we not simply call the function here (like what was done before)?

Also, would we need to invoke `exec` on line 44 [here](https://github.com/thelastpickle/cassandra-medusa/pull/313/files#diff-7e552c6093dd30777eac14ab0ab705b391b0d963d156ccda72178fef2b3514a6R44) as well, or is there no issue for restore?

emerkle826 (changes requested):
This looks ok to me. Just a couple questions about `exec` and restore that maybe do not need to be addressed.

k8s/docker-entrypoint.sh
jsanda: I need to clean this up. Provided the function call does not result in a new shell/process we should be able to leave it as is and just do the `exec` inside the `grpc` function. I will test.

>Also, would we need to invoke exec on line 44 here as well, or is there no issue for restore?

Initially I didn't think we needed it for restore since it is a long-live server process; however a restore operation could take a long time. I think we probably want it for restore as well to be on the safe side.

medusa/service/grpc/server.py
jsanda: With respect to testing, the issue is specific to containerized environments. If you run the gRPC server locally and send it `SIGTERM`, the python process will terminate. The integration tests already performa shutdown on the gRPC server, but they do not run in a containerized environment

adejanovski: gotcha 👍

k8s/docker-entrypoint.sh
adejanovski: kk, thanks

jsanda:
@adejanovski can I get another review on this?

adejanovski (approved):
LGTM 👍 
Let's wait for CI to turn green and then you can squash the commits and merge the PR.

PR #315, opened 2021-03-24
Support a flag to disable md5 comparisons during backups & Azure multipart MD5 fix

mkcello96:
This allows medusa users to optionally disable the md5 checks that Medusa does during backups (the checks will remain enabled by default). 
To explain in further detail what this flag does: during node backups to cloud storage, if a file is found in the cache AND that file is found to have the same size and MD5 hash as the one that is found in the manifest, the local file is presumed to be same as the one in the manifest, and therefore can be skipped during backup.

When this flag is specified in a backup command, the only change to the above behavior is that the MD5 has check is skipped when deciding if the local file is the same as the one in the manifest. 
I expect some medusa users will be OK with removing these MD5 checks, as SSTables are immutable.

In my experience enabling this flag sped up differential backups by more than 10x.

In addition, there is an Azure-specific fix in this PR for the multipart MD5 check. Prior to this PR,[ Azure was using a different method of computing MD5s for multipart files](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/azure_storage.py#L98), which followed AWS standards. These are not followed in Azure. Further, [az-cli only uses the common way of computing MD5s when uploading blobs to the cloud](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/azure_blobs_storage/azcli.py#L71). Therefore there was a mismatch in MD5 hashes between those in the manifest and those calculated when a file was found in the cache, causing re-uploads of all multipart files during backups.


rzvoncek (dismissed):
The changes look good to me. However, I'd still like a second pair of eyes. @adejanovski , could you please have a look once you have a moment?

adejanovski (commented):
It's great to have that PR in @mkcello96, thanks!

I have a few thoughts to share here:

- I think it would be good to make this a config setting primarily, and possibly make it possible to override it with a flag
- the `verify` command will perform hash checks as well through the `blob_matches_manifest()` method, which should be subject to the same setting (if backups are taken without checking hashes, we can assume the check should be disabled at verify time, unless requested otherwise explicitly)
- I'm wondering if maybe we could make this the default behavior, with an option to enable md5 checks. I know that hash computation takes quite a lot of time at backup time for everyone and I can think of a single case where it could be useful after restoring an old backup. If md5 checks were disabled, we could then advise to run a full backup instead of a differential one after the restore.

@rzvoncek, wdyt?

mkcello96:
@adejanovski I like all of those ideas. Just FYI, for our medusa usage we're relying on the `verify` command simply to validate MD5 hashes if needed. It's an audit/security requirement to be able to perform that check. 
But for backups there's no need for the MD5 checks. 
So our config would have MD5 checks disabled, but I can flip the boolean of the flag in this PR to be able to enable it during `verify`s (as you suggested).

mkcello96:
@adejanovski I added the changes you discussed. Let me know what you think. Happy to make changes for the wiki pages as well if needed.
Thanks!

mkcello96:
I tried to please the sonarcloud checker but I don't see a reasonable way to prevent the duplicate lines in the integration_steps file. I think it is mostly complaining that I added `"enable_md5_checks": "false"` 3 times for each of the configs defined in the tests.
@adejanovski Let me know if I should look into this more. From what I understand these sonarcloud checks are only for new code added in this PR, so the build failure here should not influence other PRs after merging.

tests/integration/features/integration_tests.feature
```diff
@@ -27,15 +27,16 @@ Feature: Integration tests
...
         Then I can see the backup named "first_backup" when I list the backups
         Then I can download the backup named "first_backup" for all tables
         Then I can download the backup named "first_backup" for "medusa.test"
         Then I can see the backup status for "first_backup" when I run the status command
         Then backup named "first_backup" has 16 files in the manifest for the "test" table in keyspace "medusa"
         Then the backup index exists
         Then the backup named "first_backup" has 2 SSTables for the "test" table in keyspace "medusa"
-        Then I can verify the backup named "first_backup" successfully
+        Then I can verify the backup named "first_backup" with md5 enabled = "false" successfully
+        Then I can verify the backup named "first_backup" with md5 enabled = "true" successfully
```
mkcello96, line 39: Alternatively, I can add 2 different tests if that is more desirable. One for just `I can verify the backup named "first_backup" `, with md5 disabled, and another for `I can verify the backup named "first_backup" with md5 enabled successfully`

tests/integration/features/integration_tests.feature
```diff
@@ -606,8 +610,8 @@ Feature: Integration tests
         When I create the "test" table with secondary index in keyspace "medusa"
         When I load 100 rows in the "medusa.test" table
         When I run a "ccm node1 nodetool flush" command
-        When I perform a backup in "differential" mode of the node named "first_backup"
-        Then I can verify the backup named "first_backup" successfully
+        When I perform a backup in "differential" mode of the node named "first_backup" with md5 enabled = "false"
+        Then I can verify the backup named "first_backup" with md5 enabled = "false" successfully
```
adejanovski, line 614:
```suggestion
        When I perform a backup in "differential" mode of the node named "first_backup" with md5 checks "disabled"
        Then I can verify the backup named "first_backup" with md5 checks "disabled" successfully
```

tests/integration/features/integration_tests.feature
```diff
@@ -650,10 +654,10 @@ Feature: Integration tests
...
         When I load 100 rows in the "medusa.test" table
         When I run a "ccm node1 nodetool flush" command
-        When I perform a backup in "full" mode of the node named "first_backup"
-        When I perform a backup in "differential" mode of the node named "second_backup"
-        Then I can verify the backup named "first_backup" successfully
-        Then I can verify the backup named "second_backup" successfully
+        When I perform a backup in "full" mode of the node named "first_backup" with md5 enabled = "false"
+        When I perform a backup in "differential" mode of the node named "second_backup" with md5 enabled = "false"
+        Then I can verify the backup named "first_backup" with md5 enabled = "false" successfully
+        Then I can verify the backup named "second_backup" with md5 enabled = "false" successfully
```
adejanovski, line 660:
```suggestion
        When I perform a backup in "full" mode of the node named "first_backup" with md5 checks "disabled"
        When I perform a backup in "differential" mode of the node named "second_backup" with md5 checks "disabled"
        Then I can verify the backup named "first_backup" with md5 checks "disabled" successfully
        Then I can verify the backup named "second_backup" with md5 checks "disabled" successfully
```

tests/integration/features/steps/integration_steps.py
```diff
@@ -598,7 +599,8 @@ def i_am_using_storage_provider_with_grpc_server(context, storage_provider, clie
     config["monitoring"] = {"monitoring_provider": "local"}
 
     config["checks"] = {
-        "health_check": "cql"
+        "health_check": "cql",
+        "enable_md5_checks": "false"
```
adejanovski, line 603: In order to avoid duplications (which indeed make Sonar unhappy), we could define a global variable at the top:
```
config_checks = {"health_check": "cql", "enable_md5_checks": "false"}
```

And then use: 

```
     config["checks"] = config_checks
```

On these lines and also on line 731.

tests/integration/features/steps/integration_steps.py
```diff
@@ -823,10 +826,11 @@ def _i_run_a_whatever_command(context, command):
     os.popen(command).read()
 
 
-@when(r'I perform a backup in "{backup_mode}" mode of the node named "{backup_name}"')
-def _i_perform_a_backup_of_the_node_named_backupname(context, backup_mode, backup_name):
+@when(r'I perform a backup in "{backup_mode}" mode of the node named'
+      r' "{backup_name}" with md5 enabled = "{md5_enabled}"')
+def _i_perform_a_backup_of_the_node_named_backupname(context, backup_mode, backup_name, md5_enabled):
```
adejanovski, line 831:
```suggestion
@when(r'I perform a backup in "{backup_mode}" mode of the node named'
      r' "{backup_name}" with md5 checks "{md5_enabled}"')
def _i_perform_a_backup_of_the_node_named_backupname(context, backup_mode, backup_name, md5_enabled):
```

tests/integration/features/steps/integration_steps.py
```diff
@@ -823,10 +826,11 @@ def _i_run_a_whatever_command(context, command):
...
     (actual_backup_duration, actual_start, end, node_backup, node_backup_cache, num_files, start) \
-        = medusa.backup_node.main(context.medusa_config, backup_name, None, backup_mode)
+        = medusa.backup_node.main(context.medusa_config, backup_name, None, md5_enabled, backup_mode)
```
adejanovski, line 833:
```suggestion
        = medusa.backup_node.main(context.medusa_config, backup_name, None, 
                                                            True if md5_enabled == "enabled" else False, backup_mode)
```

tests/integration/features/steps/integration_steps.py
```diff
@@ -987,9 +991,10 @@ def _the_backup_named_backupname_has_nb_sstables_for_the_whatever_table(
         assert len(sstables) == int(nb_sstables)
 
 
-@then(r'I can verify the backup named "{backup_name}" successfully')
-def _i_can_verify_the_backup_named_successfully(context, backup_name):
-    medusa.verify.verify(context.medusa_config, backup_name)
+@then(r'I can verify the backup named "{backup_name}" with md5 enabled = "{md5_enabled}" successfully')
+def _i_can_verify_the_backup_named_successfully(context, backup_name, md5_enabled):
+    enable_md5_checks = medusa.utils.evaluate_boolean(md5_enabled)
+    medusa.verify.verify(context.medusa_config, backup_name, enable_md5_checks)
```
adejanovski, line 997:
```suggestion
@then(r'I can verify the backup named "{backup_name}" with md5 checks "{md5_enabled}" successfully')
def _i_can_verify_the_backup_named_successfully(context, backup_name, md5_enabled):
    medusa.verify.verify(context.medusa_config, backup_name, True if md5_enabled == "enabled" else False)
```

adejanovski (changes requested):
Thanks a lot for handling this work @mkcello96, I've been trying to find someone to add this flag for a while now 😅 

I've suggested a few minor changes to the Gherkin wording to make it more natural to read and also a few recommendations on using Python ternaries rather than classic if/else structures.
Otherwise it looks great to me!

Would you, by any chance, have some numbers on the time savings that can be expected when disabling the hash checks?

tests/integration/features/steps/integration_steps.py
mkcello96: Thanks that worked :)

mkcello96:
> > @adejanovski I added the changes you discussed. Let me know what you think. Happy to make changes for the wiki pages as well if needed.
> > Thanks!
> 
> Thanks for offering to update the wiki, we're currently in the process of moving the docs back in tree in #317 (using a wiki was my mistake in the first place, it does not allow pushing new features and the associated documentation in the same PR, which is counterproductive). I think we should be able to merge that PR today, which will allow you to rebase on top of `master` and update the docs right in this PR.

Thanks for the review! I'll keep an eye on that PR and rebase after that's in.

mkcello96:
@adejanovski Readme changes are in!

adejanovski (approved):
Awesome work @mkcello96, thanks a lot!

PR #317, opened 2021-03-25
Bring back docs in tree and rebrand Medusa where needed

adejanovski:


CONTRIBUTING.md
```diff
@@ -5,7 +5,7 @@ Contributing
 
 This project adheres to the [Spotify FOSS Code of Conduct][code-of-conduct]. By participating, you are expected to honor this code.
 
-[code-of-conduct]: https://github.com/spotify/medusa/blob/master/CODE_OF_CONDUCT.md
+[code-of-conduct]: ../CODE_OF_CONDUCT.md
```
emerkle826, line 8: @adejanovski Do we want to use a _hard_ link here and put the absolute path to the CoC, or leave it relative?

docs/Performing-backups.md
```diff
@@ -0,0 +1,103 @@
...
+* The token map, a list of nodes and their token ownership.
+* The manifest, a list of backed up files with their md5 hash.
+
+## Full And Differential Backups
+
+All Medusa backups only copy new SSTables from the nodes, reducing the network traffic needed. It then has two ways of managing the files in the backup catalog that we call Full or Differential backups. For Differential backups only references to SSTables are kept by each new backup, so that only a single instance of each SStable exists no matter how many backups it is in. Differential backups are the default and in operations at Spotify reduced the backup size for some clusters by up to 80%.
+
+Full backups create a complete copy of all SSTables on the node each time they run. Files that have not changed since the last backup will be copied in the backup catalog into the new backup (and not copied off the node). In contrast to the differential method which only creates a reference to files. Full backups are useful when you need to take a complete copy and have all the files in a single location.
+
+![Full backups with Medusa for Apache Cassandra](images/medusa_current_full_backups.png)
```
emerkle826, line 97: Same comment here. Do we want hard links or relative links to the image files? Hard links are good in case we move pages around structurally. Relative links are better if everything moves and stays relative to each other.

CONTRIBUTING.md
adejanovski: I think we should keep relative paths, but it's in the same directory as the CONTRIBUTING.md file, so the link should just be `CODE_OF_CONDUCT.md`, right?

docs/Performing-backups.md
adejanovski: My preference would go to relative links.

CONTRIBUTING.md
emerkle826: You're right, not sure how I missed this....

docs/Performing-backups.md
emerkle826: :+1:

adejanovski:
It all looks good to me @emerkle826, wdyt?

emerkle826:
> It all looks good to me @emerkle826, wdyt?

Looks good to me as well :+1:

arodrime:
@adejanovski LGTM as well fwiw :). You'll need to rebase on master though it seems.

Good call on making docs "in-tree" 👏 .

Issue #241, opened 2020-12-22
Problem restoring to the same cluster - cassandra-medusa (0.8.0.dev0)


bkrajmalnik1:
Hello,

I am running Medusa on CentOS.
Single node restores are working without issue.
However, a cluster-wide restore is failing.
By checking a debug level log from medusa, I can see that on all 3 nodes the data files are moved from the backup repository to the data directory, and the commitlogs are removed.
I ran the medusa on the seed node, and did not pass an explicit seed-node parameter, so it inferred itself as the seed node.
From medusa.log, we see the following:

[2020-12-21 17:11:46,856] INFO: No --seeds specified so we will not wait for any
[2020-12-21 17:11:46,856] INFO: Starting Cassandra
[2020-12-21 17:11:46,857] DEBUG: Starting Cassandra with ['sudo', 'systemctl', 'start', 'cassandra']
[2020-12-21 17:11:46,916] DEBUG: Cleaning (/tmp/medusa-restore-e84b3ad0-16d2-4fe3-b2c2-8a3b7d516c72)
[2020-12-21 17:11:46,917] DEBUG: Remove folder /tmp/medusa-restore-e84b3ad0-16d2-4fe3-b2c2-8a3b7d516c72 and content

On the other nodes, I see the following:

[2020-12-21 17:11:47,221] DEBUG: This server has systemd: True
[2020-12-21 17:11:47,222] WARNING: is ccm : 0
[2020-12-21 17:11:47,521] DEBUG: The node cass01.testdomain.local is not up yet...
[2020-12-21 17:11:47,521] INFO: No seeds are up yet, will wait a minute
[2020-12-21 17:12:47,543] DEBUG: This server has systemd: True
[2020-12-21 17:12:47,544] WARNING: is ccm : 0
[2020-12-21 17:12:47,838] DEBUG: The node cass01.testdomain.local is not up yet...
[2020-12-21 17:12:47,838] INFO: No seeds are up yet, will wait a minute

Looking at the code, it appears that it is checking by executing this code

def is_open(host, port):
    timeout = 3
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.settimeout(timeout)
    is_accessible = False
    try:
        s.connect((host, port))
        s.shutdown(socket.SHUT_RDWR)
        is_accessible = True
    except socket.error as e:
        logging.debug('Port {} close on host {}'.format(port, host), exc_info=e)
        is_accessible = False
    finally:
[...]

PR #319, opened 2021-03-29
Remote restore w/ encrypted port checks

jeffbanks:
Includes:

- #241
-- Storage and Native - include encrypted ports in checks.
-- Exception handling w/ fix for `is_open` returning incorrect status.
- Addresses a set of Sonar warnings scoped to `cassandra_utils`.

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -15,29 +15,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import itertools
-import logging
-import time
 
 import fileinput
+import itertools
+import logging
 import pathlib
 import shlex
 import socket
 import subprocess
-import yaml
-from cassandra.auth import PlainTextAuthProvider
-from cassandra.cluster import Cluster, ExecutionProfile
-from cassandra.policies import WhiteListRoundRobinPolicy
-from cassandra.util import Version
-from retrying import retry
-from ssl import SSLContext, PROTOCOL_TLSv1_2, CERT_REQUIRED
+import time
+import os
+from medusa.utils import null_if_empty
+
 from subprocess import PIPE
 
-from medusa.host_man import HostMan
+import yaml
+
+from cassandra.cluster import Cluster, ExecutionProfile
+from cassandra.policies import WhiteListRoundRobinPolicy
+from cassandra.auth import PlainTextAuthProvider
+from ssl import SSLContext, PROTOCOL_TLS, CERT_REQUIRED
 from medusa.network.hostname_resolver import HostnameResolver
-from medusa.nodetool import Nodetool
 from medusa.service.snapshot import SnapshotService
-from medusa.utils import null_if_empty
+from medusa.nodetool import Nodetool
+from cassandra.util import Version
+from retrying import retry
+from medusa.host_man import HostMan
 
 
 class SnapshotPath(object):
@@ -66,7 +69,7 @@ class CqlSessionProvider(object):
             self._auth_provider = auth_provider
 
         if cassandra_config.certfile is not None:
-            ssl_context = SSLContext(PROTOCOL_TLSv1_2)
+            ssl_context = SSLContext(PROTOCOL_TLS)
             ssl_context.load_verify_locations(cassandra_config.certfile)
             ssl_context.verify_mode = CERT_REQUIRED
             if cassandra_config.usercert is not None and cassandra_config.userkey is not None:
@@ -80,9 +83,9 @@ class CqlSessionProvider(object):
             'local': ExecutionProfile(load_balancing_policy=load_balancing_policy)
         }
 
-    def new_session(self, retry=False):
+    def new_session(self, is_retry=False):
         """
-        Creates a new CQL session. If retry is True then attempt to create a CQL session with retry logic. The max
+        Creates a new CQL session. If is_retry is True then attempt to create a CQL session with retry logic. The max
         number of retries is currently hard coded at 5 and the delay between attempts is also hard coded at 5 sec. If
         no session can be created after the max retries is reached, an exception is raised.
          """
@@ -92,7 +95,7 @@ class CqlSessionProvider(object):
                           execution_profiles=self._execution_profiles,
                           ssl_context=self._ssl_context)
 
-        if retry:
+        if is_retry:
             max_retries = 5
             attempts = 0
 
@@ -294,14 +297,17 @@ class CassandraConfigReader(object):
                     self._config['native_transport_port'] is not None:
                 return self._config['native_transport_port']
 
-            # default encrypted
+        client_encryption = self._config['client_encryption_options']['enabled']
+        if client_encryption in self._config and client_encryption and client_encryption is True:
+            native_transport_port_ssl = self._config['native_transport_port_ssl']
+            if native_transport_port_ssl in self._config and native_transport_port_ssl:
+                return native_transport_port_ssl
             return "9142"
-        else:
-            # Condition for client encryption not enabled, default non-encrypted port.
-            if 'native_transport_port' in self._config and self._config['native_transport_port'] is not None:
-                return self._config['native_transport_port']
-            # default non-encrypted
-            return "9042"
+
+        native_transport_port = self._config['native_transport_port']
+        if native_transport_port in self._config and native_transport_port:
+            return native_transport_port
+        return "9042"
 
     @property
     def rpc_port(self):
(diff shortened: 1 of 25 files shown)


medusa/cassandra_utils.py
```diff
@@ -619,11 +632,14 @@ def is_open(host, port):
...
-        logging.debug('Port {} close on host {}'.format(port, host), exc_info=e)
-        is_accessible = False
+        logging.error('Port {} close on host {}'.format(port, host), exc_info=e)
     finally:
-        s.close()
-        return is_accessible
+        try:
+            s.close()
+        except os.error as ose:
+            logging.error('Port {} close on host {}'.format(port, host), exc_info=ose)
```
adejanovski, line 640:
```suggestion
        s.close()
```

After checking, this call cannot fail so we don't need to wrap it in a try/except block.

medusa/cassandra_utils.py
```diff
@@ -238,45 +239,57 @@ def saved_caches_directory(self):
...
         return "9160"
 
     @property
     def seeds(self):
         seeds = list()
-        if 'seed_provider' in self._config:
-            if self._config['seed_provider']:
-                if self._config['seed_provider'][0]['class_name'].endswith('SimpleSeedProvider'):
-                    return self._config.get('seed_provider')[0]['parameters'][0]['seeds'].replace(' ', '').split(',')
+        if 'seed_provider' in self._config and self._config['seed_provider'] and \
```
adejanovski, line 286: neat 👍 

I think we were afraid that unlike Java, Python would evaluate all the conditions at once, which could trigger an exception if the key was missing. Good to know it's not the case!

adejanovski (commented):
So far so good @jeffbanks 👍 
Just two minor changes required for now.

Let me know when the rest of the PR gets in with the changes on which port gets checked to see if seeds are up.

ashokkoti:
Hi @jeffbanks @adejanovski 

In the is_open method , for the s.connect port passed is of the type string instead of int . 

With type conversion of port to the existing master code able to restore (on non ssl port 9042) 


def is_open(host, port):
    timeout = 3
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.settimeout(timeout)
    is_accessible = False
    port = int(port)
    try:

jeffbanks:
@adejanovski

The most recent additions:
* Refactor logic in `is_node_up` to support c* and ccm health checks.
* Encrypt enablement verification and comments to support native and storage ports options.
* Unit test coverage that leverages c* configuration resource artifacts (as separate .yaml files)

I'm open to addressing port test inputs differently in the future as needed to customize the various encrypt communication enablement and disablement.  Since there were no unit tests for that specific logic, I felt the need to get something in place.

medusa/cassandra_utils.py
jeffbanks: Verified that this is still the case (exception for missing key) unless there is another option, I will leave this approach in place.

jeffbanks:
@adejanovski I will the integration tests (found some failures) to support recent changes in an upcoming push.  I wanted to get your eyes on the most recent health logic changes.

medusa/cassandra_utils.py
```diff
@@ -238,45 +239,61 @@ def saved_caches_directory(self):
...
+        # When setting native_transport_port_ssl, expecting that non-encrypted will still be over existing
+        # native_transport_port. The encrypted will be over the native_transport_port_ssl.
+        # When not setting an alternate native_transport_port_ssl , the encrypted will be over
+        # the existing native_transport_port
+
+        if 'client_encryption_options' in self._config and \
+                self._config['client_encryption_options']['enabled'] == "true" \
+                and 'native_transport_port_ssl' in self._config \
+                and self._config['native_transport_port_ssl']:
+            return self._config['native_transport_port_ssl']
```
adejanovski, line 272: You've removed the defaults here and in the storage port. Did you have a specific reason for this?
There's a case where the settings may not be present in the `cassandra.yaml` file and will then fallback to the defaults. Unless I'm missing something, we're not covering this case here.

My expectation would be to : 

- Check if encryption is enabled
- if so:
  - check if the SSL port is defined and return it, or return the default (9142)
- if not:
  - check if the non-SSL native port is defined and return it, or return the default (9042)

Same goes for the storage port by checking internode encryption.

medusa/cassandra_utils.py
```diff
@@ -340,15 +357,15 @@ def hostname(self):
 
     @property
     def storage_port(self):
-        return self._storage_port
+        return int(self._storage_port)
```
adejanovski, line 360: yup, good call 👍

medusa/cassandra_utils.py
```diff
@@ -514,22 +531,22 @@ def replaceTokensInCassandraYamlAndDisableBootstrap(self, token_list):
         initial_token_line_found = False
         auto_bootstrap_line_found = False
         for line in fileinput.input(self._cassandra_config_file, inplace=True):
-            if (line.startswith("initial_token:")):
+            if line.startswith("initial_token:"):
```
adejanovski, line 534: Getting used to Go style? 😉 

Just kidding, wfm.

medusa/cassandra_utils.py
```diff
@@ -558,38 +575,66 @@ def wait_for_node_to_go_down(config, host):
...
-            return is_cassandra_up(host, rpc_port) and is_cassandra_up(host, native_port)
+        elif check_type == 'all':
+            # Port checks to include all of: rpc, native, and storage(gossip) health.
+            return is_cassandra_up(host, rpc_port) and is_cassandra_up(host, native_port) and \
+                is_cassandra_up(host, storage_port)
         else:
-            # cql only
-            return is_cassandra_up(host, native_port)
+            # Default port checks for native OR storage(gossip).
+            return is_cassandra_up(host, native_port) or is_cassandra_up(host, storage_port)
```
adejanovski, line 632: LGTM 👍

medusa/cassandra_utils.py
```diff
@@ -654,6 +699,27 @@ class CassandraNodeNotDownError(Exception):
...
+    """
+    Raised when we detect an invalid Cassandra configuration
+    """
+
+    def __init(self, host):
+        msg = 'Cassandra config is detected with invalid settings'
+        super(Exception, self).__init__(msg)
+
+
+class CassandraCqlSessionException(Exception):
```
adejanovski, line 718: Nice addition 👍

medusa/cassandra_utils.py
```diff
@@ -654,6 +699,27 @@ class CassandraNodeNotDownError(Exception):
     Raised when we give up waiting on a node to go down, meaning nodetool statusbinary and/or statusthrift keep
     reporting open ports.
     """
+
     def __init(self, host):
         msg = 'Cassandra node {} is still up...'.format(host)
         super(CassandraNodeNotDownError, self).__init__(msg)
+
+
+class CassandraConfigInvalidError(Exception):
```
adejanovski, line 708: This one might not be needed depending on your answer to my previous comment.

tests/cassandra_utils_test.py
```diff
@@ -407,6 +408,166 @@ def test_parsing_custom_seed_provider(self):
...
+            cassandra=_namedtuple_from_dict(CassandraConfig, config['cassandra']),
+            grpc=_namedtuple_from_dict(GrpcConfig, config['grpc']),
+            kubernetes=_namedtuple_from_dict(KubernetesConfig, config['kubernetes']),
+            ssh=None,
+            checks=None,
+            logging=None,
+        )
+
+        self.assertRaises(CassandraConfigInvalidError, Cassandra, medusa_config)
+
```
adejanovski, line 570: Awesome sauce 💯 

We may have to drop the last test as valid configurations can avoid defining the ports and fallback to defaults.

medusa/cassandra_utils.py
jeffbanks: Gotcha, I see your point here.  I was thinking that if not specified, we were requiring a port designation in the c* config to avoid a health check on a port they removed.  I will remove the exception raised for that condition and use the default port literal values.

medusa/cassandra_utils.py
jeffbanks: I believe this is the source of the issues in the integration tests failing in this PR. I will refactor those to assist @ivanmp91 as well with his #320.  I had requested he do the same, and I see some integration test failures there that may be working with string types still for ports.

jeffbanks:
> Hi @jeffbanks @adejanovski
> 
> In the is_open method , for the s.connect, port passed is of the type string instead of int .
> 
> Was able to restore with type conversion of port to the existing master code (for non ssl port's : 9042)
> 
> ```
> def is_open(host, port):
>     timeout = 3
>     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>     s.settimeout(timeout)
>     is_accessible = False
>     port = int(port)
>     try:
> ```

@ashokkoti thank you for reporting the typing issue.  @ivanmp91 is working this in #320 that was previously opened.  It should be upcoming in the near future.

medusa/cassandra_utils.py
jeffbanks: Added an integration test fix for this that takes care of the other PR issues as well.  I'll push that to #320 for @ivanmp91.

jeffbanks:
Resolving as efforts in #344 scope of changes.

PR #320, opened 2021-03-30
Force int data type for healthcheck port

ivanmp91:
The socket test check https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/cassandra_utils.py#L618 requires an integer data type for port parameter, but yaml load returns a string https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/cassandra_utils.py#L214 so port healthcheck will always fail for that reason, we need to convert to integer before. That's the error that I got debugging this issue that affects while performing a full cluster restore:

```
    s.connect((host, port))
TypeError: an integer is required (got type str)
```

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -340,15 +340,15 @@ class Cassandra(object):
 
     @property
     def storage_port(self):
-        return self._storage_port
+        return int(self._storage_port)
 
     @property
     def native_port(self):
-        return self._native_port
+        return int(self._native_port)
 
     @property
     def rpc_port(self):
-        return self._rpc_port
+        return int(self._rpc_port)
 
     class Snapshot(object):
         def __init__(self, parent, tag):


jeffbanks (changes requested):
@ivanmp91, thank you for identifying this one!  I was looking in the same location for a different PR recently. Wondering if this would make sense for your PR.

My understanding is there is a `C* Reader` and `C* object`,  would it make more sense for us to add this to the C* object ([here](https://github.com/thelastpickle/cassandra-medusa/blob/44748a1aa1a4d1dd6936c7e7f88a6e1f8016e4c1/medusa/cassandra_utils.py#L341)) so we can get this for all 3 port values in one location?

**Example**

```python
@property
def storage_port(self):
    return int(self._storage_port)

@property
def native_port(self):
    return int(self._native_port)

@property
def rpc_port(self):
    return int(self._rpc_port)
```

ivanmp91:
@jeffbanks Thank you for your response! I think makes completely sense changing the data type in the class attributes, I've just updated the PR to reflect those changes. Let me know if there's anything else :)

jeffbanks:
@ivanmp91 keep an eye on #319, once fully reviewed, this should fix the integration test failures we get here.

[New addition](https://github.com/thelastpickle/cassandra-medusa/blob/b6004617d37292c11184bcde95f997ae4adedb76/medusa/restore_node.py#L205) to `medusa/restore_node.py` to handle (from `int` to `str` ) for sstableloader_args usage.

```python
if storage_port != 7000:
    sstableloader_args.append("--storage-port")
    sstableloader_args.append(str(storage_port))
...
```

adejanovski:
Merged in #327. Thanks @ivanmp91 👍

PR #323, opened 2021-04-06
Fix deprecation warnings

rhardouin:
Fix the following warnings:

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```

```
DeprecationWarning: encodestring() is a deprecated alias since 3.1, use encodebytes()
```


adejanovski (approved):
Thanks for this PR @rhardouin, I hate warnings 😉

PR #324, opened 2021-04-06
Fix unit tests

rhardouin:
- `./run_integration_tests.sh` can't be run because the shebang is not on the first line
- there is no `generate_md5_hash` symbol in `backup_node`


adejanovski (approved):
LGTM, thanks!

PR #326, opened 2021-04-07
Bypass the restore in GRPC/k8s mode if the backup doesn't exist

adejanovski:
Fixes https://github.com/k8ssandra/k8ssandra/issues/461

### DIFF
--- a/medusa/service/grpc/restore.py
+++ b/medusa/service/grpc/restore.py
@@ -21,6 +21,7 @@ from pathlib import Path
 
 import medusa.config
 import medusa.restore_node
+import medusa.listing
 
 
 def create_config(config_file_path):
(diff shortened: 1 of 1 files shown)


adejanovski:
@jsanda, I've tested it locally using this image: 
```
image:
    repository: docker.io/adejanovski/medusa
    tag: restore-fix
```


I'll create an integration test for this bug in the k8ssandra project.

medusa/service/grpc/restore.py
```diff
@@ -65,9 +66,17 @@ def configure_console_logging(config):
...
+backup_found = False
+# Checking if the backup exists for the node we're restoring.
+# Skipping restore if it doesn't exist.
+for cluster_backup in cluster_backups:
+    if cluster_backup.name == backup_name:
+        backup_found = True
+        logging.info("Starting restore of backup {}".format(backup_name))
+        medusa.restore_node.restore_node(config, tmp_dir, backup_name, in_place, keep_auth,
+                                         seeds, verify, keyspaces, tables, use_sstableloader)
+        logging.info("Finished restore of backup {}".format(backup_name))
```
jsanda, line 79: I know that backup names are unique which means we should only get into this `if` block once. Would it good to add a `break` statement at the end after line 79?

jsanda (commented):
Thanks!

medusa/service/grpc/restore.py
adejanovski: oooh yeah, most definitely 👍

adejanovski: done

jsanda (approved):

PR #353, opened 2021-05-17
Use custom config file when restoring data

### LINKED ISSUE #352: Custom config file path is ignored when restoring data
`--config-file` is ignored when restoring data because it's not generated in [_build_restore_cmd()](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/restore_cluster.py#L352-L354)

rhardouin:
Close #352

To persist a custom (i.e. non default) configuration file path, a new attribute is added to `MedusaConfig`. It's not a section like the others but it allows to be self contained.
A `file_path` set to `None` means default configuration file path so we can skip the whole `--config-file ...` part when building restore command.

Unit tests added to test both default and custom configuration file path in command line.
Existing tests migrated to pytest.

### DIFF
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -56,7 +56,10 @@ MonitoringConfig = collections.namedtuple(
 
 MedusaConfig = collections.namedtuple(
     'MedusaConfig',
-    ['storage', 'cassandra', 'ssh', 'checks', 'monitoring', 'logging', 'grpc', 'kubernetes']
+    [
+        'file_path',  # Store a specific config file path, None for default file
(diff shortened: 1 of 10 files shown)


adejanovski:
I love the idea of overriding the config this way. I was thinking of something similar to allow remote restores while not sharing the same config as the original cluster (which could lead to troubles if someone starts to backup the new cluster in the same bucket).
I'll run some tests to check that everything works nicely before approving the PR.
Thanks!

rhardouin:
Thanks for the review and for the tests to come! :+1:

adejanovski (approved):
Successfully tested ✅ 

Cool stuff @rhardouin, thanks!

PR #355, opened 2021-05-18
fix hostname_resolver.resolve_fqdn when resolve_ip_addresses is False

optplx:
During a restore-cluster, when `resolve_ip_addresses` is set to `False`, an empty string is passed to the hostname_resolver which returns this empty string instead of the IP address.

This causes the restore to fail, since sstableloader has then no argument to pass 
```
/opt/cassandra/default/bin/sstableloader', '-d', '',
```

Also I make some parameters mandatory to make sure that medusa behaves as configured (all calls in the codebase do have  those options so it should not break the current behaviors)

### DIFF
--- a/medusa/network/hostname_resolver.py
+++ b/medusa/network/hostname_resolver.py
@@ -17,16 +17,17 @@ import socket
 import logging
 
 
-class HostnameResolver():
-    def __init__(self, resolve_addresses=True):
+class HostnameResolver:
+    def __init__(self, resolve_addresses):
         self.resolve_addresses = resolve_addresses
-        logging.debug
 
     def resolve_fqdn(self, ip_address=''):
-        if str(self.resolve_addresses) == "False":
-            logging.debug("Not resolving {} as requested".format(ip_address))
-            return ip_address
+        ip_address_to_resolve = ip_address if ip_address != '' else socket.gethostbyname(socket.getfqdn())
 
-        fqdn = socket.getfqdn(ip_address)
-        logging.debug("Resolved {} to {}".format(ip_address, fqdn))
+        if str(self.resolve_addresses) == "False":
+            logging.debug("Not resolving {} as requested".format(ip_address_to_resolve))
+            return ip_address_to_resolve
+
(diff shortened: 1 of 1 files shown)


medusa/restore_node.py
```diff
@@ -195,7 +196,7 @@ def invoke_sstableloader(config, download_dir, keep_auth, fqtns_to_restore, stor
                     cql_username = 'foo' if config.cassandra.cql_username is None else config.cassandra.cql_username
                     cql_password = 'foo' if config.cassandra.cql_password is None else config.cassandra.cql_password
                     sstableloader_args = [config.cassandra.sstableloader_bin,
-                                          '-d', hostname_resolver.resolve_fqdn() if cassandra_is_ccm == 0
+                                          '-d', hostname_resolver.resolve_fqdn(socket.gethostbyname(socket.getfqdn())) if cassandra_is_ccm == 0
```
adejanovski, line 199: Maybe we could avoid some duplication by coding this logic into `hostname_resolver.resolve_fqdn()` directly?

We could revert to making the ip_address optional again (and empty by default) and then use `socket.gethostbyname(socket.getfqdn())` if it's indeed empty in the method implementation.

it would give something like: 

```
    def resolve_fqdn(self, ip_address=''):
        ip_address_to_resolve = ip_address if ip_address != '' else socket.gethostbyname(socket.getfqdn())
        if str(self.resolve_addresses) == "False":
            logging.debug("Not resolving {} as requested".format(ip_address_to_resolve))
            return ip_address_to_resolve
        fqdn = socket.getfqdn(ip_address_to_resolve)
        logging.debug("Resolved {} to {}".format(ip_address_to_resolve, fqdn))
        return fqdn
```

wdyt?

optplx: Why not yes, to be honest I found this `hostname_resolver.resolve_fqdn(socket.gethostbyname(socket.getfqdn()))` not really nice :p

optplx: @adejanovski fix pushed 👌

adejanovski (approved):

PR #358, opened 2021-05-20
add codecov config

adejanovski:
This PR will ensure that codecov won't try to evaluate coverage until all parallel CI runs are done, which would yield premature failures in the checks.
Also, since external PRs can't run the cloud integration tests, their coverage is reduced to ~64%. This change will allow the codecov check so succeed at 63% of coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,15 +2,8 @@ coverage:
   precision: 2
   round: down
   range: "63...100"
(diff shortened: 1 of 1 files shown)


jeffbanks (approved):
Looks good, will help with the `codecov` noise on the other PRs.

PR #361, opened 2021-06-02
backup-cluster default name with timestamp

optplx:
This makes the `backup-name` optional in the `backup-cluster` command.
I have followed what is done for the `backup-node` concerning the naming scheme

### DIFF
--- a/medusa/backup_cluster.py
+++ b/medusa/backup_cluster.py
@@ -28,9 +28,10 @@ from medusa.storage import Storage
 from medusa.network.hostname_resolver import HostnameResolver
 
 
-def orchestrate(config, backup_name, seed_target, stagger, enable_md5_checks, mode, temp_dir,
+def orchestrate(config, backup_name_arg, seed_target, stagger, enable_md5_checks, mode, temp_dir,
                 parallel_snapshots, parallel_uploads):
     backup = None
+    backup_name = backup_name_arg or datetime.datetime.now().strftime('%Y%m%d%H')
     monitoring = Monitoring(config=config.monitoring)
     try:
         backup_start_time = datetime.datetime.now()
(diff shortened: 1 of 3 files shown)


medusa/backup_cluster.py
```diff
@@ -28,9 +28,10 @@
 from medusa.network.hostname_resolver import HostnameResolver
 
 
-def orchestrate(config, backup_name, seed_target, stagger, enable_md5_checks, mode, temp_dir,
+def orchestrate(config, backup_name_arg, seed_target, stagger, enable_md5_checks, mode, temp_dir,
                 parallel_snapshots, parallel_uploads):
     backup = None
+    backup_name = backup_name_arg or datetime.datetime.now().strftime('%Y%m%d%H')
```
adejanovski, line 34: I think we should go down to the minute resolution. Folks may want to perform more than one backup an hour (especially when testing).

medusa/medusacli.py
```diff
@@ -131,7 +131,7 @@ def backup(medusaconfig, backup_name, stagger, enable_md5_checks, mode):
 
 
 @cli.command(name='backup-cluster')
-@click.option('--backup-name', help='Backup name', required=True)
+@click.option('--backup-name', help='Backup name of the backup, defaults to current local timestamp')
```
adejanovski, line 134: It's not really going to be a "timestamp" but rather a datetime.

adejanovski (changes requested):
Cool addition 👍 
I have two small comments and need to run some tests once they're addressed.

Thanks!

medusa/backup_cluster.py
optplx: That's also what I did initially but then I found [this](https://github.com/thelastpickle/cassandra-medusa/blob/a00a45593c44b3e4f99adaa1a89475af987c1a9b/medusa/backup_node.py#L161) so I thought it would be good to be consistent. Imo I would prefer to go down to the minute indeed, wdyt?

adejanovski: Yes, I think we should go down to the minute in both places. Would you mind changing the backup-node code to use a minute resolution as well as part of this PR? It would be much appreciated!

adejanovski (approved):

PR #366, opened 2021-06-09
Add support for Azure backend in K8ssandra docker image

adutra:


adejanovski (approved):

k8s/Dockerfile
```diff
@@ -46,7 +49,7 @@ RUN groupadd -r cassandra --gid=999 && useradd -r -g cassandra --uid=999 --creat
 COPY --from=base /install /usr/local
 
 # wget could happen in the build-phase
-RUN apt-get update && apt-get install -y python3 wget
+RUN apt-get update && apt-get install -y python3 python3-setuptools wget
```
emerkle826, line 52: You could add `&& rm -rf /var/lib/apt/lists/*` to the end of this line to clean up the Apt cache, but it only saves 30-40 MB of space. I know you are looking to reduce the image size more than that, but every little bit helps.

emerkle826 (dismissed):

k8s/Dockerfile
adutra: You are right I was able to save about 30 MB. Thanks for the tip!

adutra:
So, I was able to reduce the image size to 1.33 GB to 716.35 MB \o/

Implemented in 1baee54.

BUT: this is done at the cost of only installing a few, absolutely required azure modules. **I don't know if this is future-proof.**

A bit of context: azure-cli has 89 dependencies on other azure PyPI packages + their own transitive dependencies.

I was able to reduce that list to just:

```
azure-cli-core
azure-mgmt-storage
packaging
```

And on top of that, I install the umbrella `azure-cli` package with the `--no-deps` flag (because it provides the `az` command).

emerkle826 (dismissed):

emerkle826:
LGTM. Just need to rebase before merging.

adejanovski (approved):

Issue #372, opened 2021-06-18
Backup to s3 us-east-1


optplx:
Hello, 
according to https://github.com/apache/libcloud/blob/trunk/libcloud/storage/types.py#L53, the storage_provider for the AWS us-east-1 region should be `s3`. However I have an issue when using this value in my config files:

```
 ERROR: This error happened during the cluster backup: Invalid or unsupported region: default
Traceback (most recent call last):
File "/usr/local/lib/python3.7/dist-packages/medusa/backup_cluster.py", line 50, in orchestrate
storage = Storage(config=config.storage)
File "/usr/local/lib/python3.7/dist-packages/medusa/storage/__init__.py", line 72, in __init__
self.storage_driver = self._connect_storage()
File "/usr/local/lib/python3.7/dist-packages/medusa/storage/__init__.py", line 92, in _connect_storage
s3_storage = S3Storage(self._config)
File "/usr/local/lib/python3.7/dist-packages/medusa/storage/abstract_storage.py", line 39, in __init__
self.driver = self.connect_storage()
File "/usr/local/lib/python3.7/dist-packages/medusa/storage/s3_storage.py", line 102, in connect_storage
aws_access_key_id, aws_secret_access_key, token=aws_security_token, region=region
File "/usr/local/lib/python3.7/dist-packages/libcloud/storage/drivers/s3.py", line 1134, in __init__
raise ValueError('Invalid or unsupported region: %s' % (region))
ValueError: Invalid or unsupported region: default
```
This comes from the region being set here https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/config.py#L111

It looks like this variable was not meant to be used this way, is it the correct way to go for the `us-east-1` region on AWS? Please note that for other regions we do not need this `region` parameter, it is discovered from the `storage_provider` instead

optplx:
Adding some more details after digging, the difference between us-east-1 and the other regions is here https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/s3_storage.py#L99

For us-east-1, we do not override the region and we end up with the one from the config set to `default`.

I will open a PR to discuss the potential solutions here

optplx:
as discussed in the PR, the new way of doing is by configuring as follow:
```
storage_provider = s3
region = <my_region>
```

The form `storage_provider = s3_<region_name>`  is actually the old way

PR #373, opened 2021-06-21
fix s3 storage on us-east-1 region

optplx:
Fixes https://github.com/thelastpickle/cassandra-medusa/issues/372

I may miss something here but I see no reason to have a special treatment for the `Provider.S3` case (corresponding to the `us-east-1` region.

The `get_driver` function returns the correct region if we remove the condition  L99 in `medusa/storage/s3_storage.py `
```
>>> get_driver('s3').region_name
'us-east-1'
```

### DIFF
--- a/medusa/storage/s3_storage.py
+++ b/medusa/storage/s3_storage.py
@@ -95,9 +95,7 @@ class S3Storage(S3BaseStorage):
             raise NotImplementedError("No valid method of AWS authentication provided.")
 
         cls = get_driver(Provider.S3)
-        region = self.config.region
-        if self.config.storage_provider != Provider.S3:
-            region = get_driver(self.config.storage_provider).region_name
+        region = get_driver(self.config.storage_provider).region_name
         driver = cls(
             aws_access_key_id, aws_secret_access_key, token=aws_security_token, region=region
         )


adejanovski:
The `s3` storage provider used in conjunction with a `region` setting is the new way of using libcloud to interact with S3. The "old" way used storage providers that were specific to a region, which is not very dynamic as the code had to be updated each time a new AWS region was made available.
We recently rewrote the S3 classes to allow using any S3 compatible object storage system, which relies on a region being set (and it's not just for AWS S3, which is why "default" is there).
For backwards compatibility reasons, we need to set the region using the driver for region specific storage providers (such as `s3_us_west_oregon` for example), which was done with the lines you deleted if I'm correct :) 

Your change currently makes it so that the `s3` provider can't be used with any other region, which is the opposite of what libcloud is trying to do in making it the provider for all regions (us-east-1 being the default region unless specified otherwise).

optplx:
I see, I thought that this was just an outdated way of referring to `us-east-1`  since it is the oldest region of AWS if I am not mistaken.
If I understand correctly, I need to change all my configs with the below  then
```
storage_provider = s3
region = <my_region>
```

I will give it a try, thanks !

PR #378, opened 2021-06-28
Placement awareness - dc and rack

jeffbanks:
Includes effort to introduce `dc` and `rack` into the CqlSession `tokenmap` accessible from c* utility.

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -147,8 +147,8 @@ class CqlSession(object):
                 return token.value
         raise RuntimeError('Unable to get current token')
 
-    def datacenter(self):
-        logging.debug('Checking datacenter...')
+    def placement(self):
+        logging.debug('Checking placement using dc and rack...')
         listen_address = socket.gethostbyname(self.cluster.contact_points[0])
         token_map = self.cluster.metadata.token_map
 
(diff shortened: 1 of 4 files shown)


adejanovski (commented):
LGTM! I'll run some tests to check the behavior and backwards compatibility.

medusa/cassandra_utils.py
```diff
@@ -183,10 +183,13 @@ def get_token(host_token_pair):
...
                 'tokens': tokens,
-                'is_up': host.is_up
+                'is_up': host.is_up,
+                'rack': host.rack,
+                'dc': host.datacenter
             }
             for host, tokens in host_tokens_pairs
-            if host.datacenter == datacenter
+            if host.datacenter == dc_rack_pair[0]
+            if host.rack == dc_rack_pair[1]
```
adejanovski, line 192: This is restricting the tokenmap to contain only a single dc/rack pair. It's one of the problems (feature) we currently have with Medusa, which restricts all operations to the local DC.

I'd rather remove the `if host.rack == dc_rack_pair[1]` line for now, which will keep the existing behavior. We can evolve this behavior and give more flexibility in the future, as it's outside of the scope of this PR.

I gave a try at the current state of this PR and on a 6 nodes cluster spreading over 2 DCs and three racks, the backup-cluster command only backed up a single node. Removing the line will correctly backup all 3 nodes in the local DC.

adejanovski (changes requested):
I just need one line to be removed but other than that it works exactly as expected 👌 
Tokenmaps are generated with the required additional fields:
Capture d’écran 2021-06-29 à 18 47 54

And I was able to successfully restore a backup taken in a previous version which didn't have the dc and rack fields.

Great stuff @jeffbanks 👍

medusa/cassandra_utils.py
jeffbanks: Sounds reasonable that we back this constraint off to support backwards compatibility.

jeffbanks:
Added a small update to docs to mention placement introduced.

adejanovski (approved):
Looks good to me now @jeffbanks, thanks!
You can merge the PR.

jeffbanks:
@adejanovski rebased w/ latest.

adejanovski (approved):

PR #379, opened 2021-06-30
update s3 storage_provider documentation

optplx:
Following https://github.com/thelastpickle/cassandra-medusa/pull/373#issuecomment-867385793,  `The s3 storage provider used in conjunction with a region setting is the new way of using libcloud to interact with S3` 

It would probably be good to change the doc, wdyt?


adejanovski (approved):
Thanks @optplx, that's going to avoid confusion. Hidden settings are the worst 😆

PR #380, opened 2021-06-30
make medusa compatible with imdsv2 on AWS

optplx:
For security reasons, we will enforce imdsv2 on our instances (see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html)

This is compatible with both v1 and v2 without having to worry about it


optplx:
hi @adejanovski do you think this is acceptable?

medusa/storage/s3_storage.py
```diff
@@ -82,7 +89,7 @@ def connect_storage(self):
                 logging.debug('Reading AWS credentials from IAM Role: %s', aws_instance_profile.text)
                 url = "http://169.254.169.254/latest/meta-data/iam/security-credentials/" + aws_instance_profile.text
                 try:
-                    auth_data = requests.get(url).json()
+                    auth_data = requests.get(url, headers=self.imds_headers).json()
```
adejanovski, line 92: Do you know if this has an impact with IMDSv1 usage of IAM roles (if that's even a thing)?
ie: is this change transparent to any existing user relying on IAM roles?

optplx: yes as per my tests: if imdsv2 is available it will use it, if not it uses imdsv1 and it is transparent

optplx: e.g with `headers=None`: 
```
requests.get("http://169.254.169.254/latest/meta-data/iam/security-credentials/xxxxx", headers=None).text
'{\n  "Code" : "Success",\n  "LastUpdated" : "2021-09-07T08:02:16Z",\n  "Type" : "AWS-HMAC",\n  "AccessKeyId" : "xxxx",\n  "SecretAccessKey" : "xxxxxxx",\n  "Token" : "xxxx==",\n  "Expiration" : "2021-09-07T14:20:41Z"\n}'
```

adejanovski: Sounds good, thanks!

adejanovski (approved):

adejanovski:
I should've seen there were 6 commits before merging... My bad 😓

optplx:
> I should've seen there were 6 commits before merging... My bad 😓

yes I tested the "rebase" feature from the console quite a lot, sorry :p

Issue #389, opened 2021-08-16
Backup method in gRPC service does not support full backups


jsanda:
The backup type is hard coded to `differential` in `server.py`. The request for this change originated from https://github.com/k8ssandra/k8ssandra/issues/1024. We already have a PR with the necessary changes in medusa-operator, https://github.com/k8ssandra/medusa-operator/pull/60.

adejanovski:
This was handled already and the grpc server now supports both modes.

PR #391, opened 2021-08-17
Make use of the backup mode passed in protobuf message

MattFellows:
This is an attempt at solving
https://github.com/thelastpickle/cassandra-medusa/issues/389

Criticisms and comments gladly recieved...

### DIFF
--- a/medusa/service/grpc/server.py
+++ b/medusa/service/grpc/server.py
@@ -34,6 +34,8 @@ from medusa.service.grpc import medusa_pb2_grpc
 from medusa.storage import Storage
 
 TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S'
+BACKUP_MODE_DIFFERENTIAL = "differential"
+BACKUP_MODE_FULL = "full"
 
 
 class MedusaService(medusa_pb2_grpc.MedusaServicer):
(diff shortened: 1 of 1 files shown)


jsanda (approved):
Thanks for the PR! Sorry it took so long for the review.

medusa/service/grpc/server.py
```diff
@@ -44,11 +46,14 @@ def __init__(self, config):
         self.storage = Storage(config=self.config.storage)
 
     def Backup(self, request, context):
-        logging.info("Performing backup {}".format(request.name))
+        logging.info("Performing backup {} (type={})".format(request.name, request.mode))
         resp = medusa_pb2.BackupResponse()
-        # TODO pass the staggered and mode args
+        # TODO pass the staggered arg
+        mode = BACKUP_MODE_DIFFERENTIAL
+        if medusa_pb2.Mode.FULL == request.mode:
```
jsanda, line 53: @MattFellows I missed this on my initial review. `medusa_pb2.Mode.FULL` is wrong. It should be `medusa_pb2.BackupRequest.Mode.FULL`. It is causing some integration tests to fail.

jsanda (changes requested):

medusa/service/grpc/server.py
MattFellows: There are 3 ways of referencing that enum that seem to pass local testing - you could of just told me that the 3rd was was wrong  (always nice to get a Monty Hall Problem reference in ;) )

jsanda:
@emerkle826 can you take a look at these two integration test failures? I am not sure why they are failing.

emerkle826:
The tests that are failing are ones that run against Cassandra trunk. The tests pass against 4.0.0 GA, so it's something in trunk that's been recently added. I'll dig in to see what's going on and figure out what needs to be changed.

emerkle826:
The test failure is due to a non-backward compatible change in some internal Cassandra APIs that Management API uses. It's not an issue with Medusa, so I think this PR can be merged and ignore the test failures for now. The fix will have to be in the Management API project [here](https://github.com/k8ssandra/management-api-for-apache-cassandra), and I'll create a ticket over there to address it.

emerkle826:
For the record, the Management API issue is being tracked [here](https://github.com/k8ssandra/management-api-for-apache-cassandra/issues/122)

jsanda:
@emerkle826 thank you for following up this. Do you think we are good to merge since the failures are unrelated to the changes in the PR?

emerkle826 (approved):

Issue #390, opened 2021-08-16
java.io.IOException: Snapshot medusa-<....> already exists.


kiddom-kq:
I've been working on standing up Medusa the last few days. Most of the issues that I hit have workarounds documented:

- https://github.com/thelastpickle/cassandra-medusa/issues/212
- https://github.com/thelastpickle/cassandra-medusa/issues/144

But this one didn't; i'm creating this issue to document the problem for future benefit.

-----

On an `admin` node, i would run:

```
$ medusa backup-cluster ...
```

Which would then come back with a failure message that appeared to be the same on each node in the cluster

When looking at the `/tmp/medusa-job-<...>/stderr` file on each node, I would see something like this:

```shell
<...>
[2021-08-11 21:56:17,250] DEBUG: https://s3-us-east-2.amazonaws.com:443 "HEAD /some-bucket-name-here/ClusterNameHere/ip-10-69-42-1.us-east-2.compute.internal/None/meta/differential HTTP/1.1" 404 0
[2021-08-11 21:56:17,251] DEBUG: [Storage] Getting object ClusterNameHere/ip-10-69-42-1.us-east-2.compute.internal/None/meta/incremental
[2021-08-11 21:56:17,255] DEBUG: https://s3-us-east-2.amazonaws.com:443 "HEAD /some-bucket-name-here HTTP/1.1" 200 0
[2021-08-11 21:56:17,267] DEBUG: https://s3-us-east-2.amazonaws.com:443 "HEAD /some-bucket-name-here/ClusterNameHere/ip-10-69-42-1.us-east-2.compute.internal/None/meta/incremental HTTP/1.1" 404 0
[2021-08-11 21:56:17,267] INFO: Node ip-10-69-42-1.us-east-2.compute.internal does not have latest backup
[2021-08-11 21:56:17,267] INFO: Starting backup
[2021-08-11 21:56:17,268] INFO: Creating snapshot
[2021-08-11 21:56:17,268] DEBUG: Executing: nodetool -u someUserNameHere -pw somePasswordStringHere snapshot -t medusa-202108112127
error: Snapshot medusa-202108112127 already exists.
-- StackTrace --
java.io.IOException: Snapshot medusa-202108112127 already exists.
```

A bizarre error because just seconds before the error showed up, the `nodetool snapshot` command was executed cluster-wide.
[...]

PR #395, opened 2021-08-24
[WIP] Sanity checking around backing up files

kiddom-kq:
While trying to deploy Medusa, i ran into a few issues. The root cause of  #390 was an issue file systems permissions and  the silent-failure property of python's `glob()`

This PR implements some of the debug logging that I wish I had had while troubleshooting and a basic sanity check to abort execution as soon as an error is observed rather than waiting for a (misleading) failure at a later point in execution.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1398) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1398
┆priority: Medium

### DIFF
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -326,7 +326,7 @@ def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot
         if _num_dirs < 1:
             _e = "Could not identify any directories where snapshot files should reside."
             logging.critical(_e)
-            raise Exception(_e)
+            raise ValueError(_e)
 
         for snapshot_path in snapshot.find_dirs():
             logging.debug("Backing up {}".format(snapshot_path))
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -323,10 +323,7 @@ def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot
         # Consider find_dirs() returning nothing to be a critical failure. At no point should a cassandra snapshot
         #   exist in 0 directories on the host.
         # We can't backup what we can't find and we consider any result other than a successful backup to be a failure
-        if _num_dirs < 1:
-            _e = "Could not identify any directories where snapshot files should reside."
-            logging.critical(_e)
-            raise ValueError(_e)
+        assert _num_dirs >= 1, "Could not identify any directories where snapshot files should reside."
 
         for snapshot_path in snapshot.find_dirs():
             logging.debug("Backing up {}".format(snapshot_path))
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -313,6 +313,21 @@ def update_monitoring(actual_backup_duration, backup_name, monitoring, node_back
 def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot):
     try:
         num_files = 0
+        _snapshot_file_dirs = snapshot.find_dirs()
+        _num_dirs = len(_snapshot_file_dirs)
+        logging.info("Found '{num_dir}' directories that may contain snapshot files.".format(num_dir=_num_dirs))
+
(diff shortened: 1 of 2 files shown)


kiddom-kq:
Would `ValueError` be more appropriate than the base `Exception` that's being raised? I didn't see too many custom exception types in the Medusa code base and of the few, none really fit this case. There are several places in the code bases where `raise Exception(error_msg)` is used so i'm not sure if there's some lint-disable comment i can leave there to prevent SonarCloud from objecting.

adejanovski:
@kiddom-kq, it looks like this broke the integration tests. You can run them locally using `./run_integration_tests.sh`, with a prereq to have ccm installed (`pip install ccm`).
Let me know if you need some assistance to debug this.

adejanovski:
> Would `ValueError` be more appropriate than the base `Exception` that's being raised? I didn't see too many custom exception types in the Medusa code base and of the few, none really fit this case. There are several places in the code bases where `raise Exception(error_msg)` is used so i'm not sure if there's some lint-disable comment i can leave there to prevent SonarCloud from objecting.

Any improvement getting a more specific use of exceptions is welcome. Feel free to make any change necessary.

kiddom-kq:
@adejanovski 

I am having some trouble understanding the test.

```shell
 ./run_integration_tests.sh --test=16  --cassandra-version=2.2.19 -vv
```

That returns a failure because the `assert` statement is thrown.

What is supposed to happen when [`find_dirs()`](https://github.com/thelastpickle/cassandra-medusa/blob/5cb39834e7f8ad1c48641af1f1548898bacd5b51/medusa/backup_node.py#L316) has no dirs?

Looking at the [test](https://github.com/thelastpickle/cassandra-medusa/blob/master/tests/integration/features/integration_tests.feature#L701):

```
    Scenario Outline: Perform a differential backup over gRPC , verify its index, then delete it over gRPC with Jolokia
        Given I have a fresh ccm cluster with jolokia "<client encryption>" running named "scenario16"
        Given I am using "<storage>" as storage provider in ccm cluster "<client encryption>" with gRPC server
        Then the gRPC server is up
        When I create the "test" table in keyspace "medusa"
        When I load 100 rows in the "medusa.test" table
        When I run a "ccm node1 nodetool flush" command
        When I perform a backup over gRPC in "differential" mode of the node named "grpc_backup"
        Then the backup index exists
        Then I verify over gRPC that the backup "grpc_backup" exists
        Then I can see the backup index entry for "grpc_backup"
        Then I can see the latest backup for "127.0.0.1" being called "grpc_backup"
        Then I delete the backup "grpc_backup" over gRPC
        Then I verify over gRPC the backup "grpc_backup" does not exist
        Then I shutdown the gRPC server
```

It looks like `When I perform a backup over gRPC in "differential" mode of the node named "grpc_backup"` is the *first* time that a backup is taken of the `scenario16` cluster. If that is correct, how can a table with 100 rows in it have no directories that need to be backed up?

When I disable the assert, the execution continues to call [`add_backup_finish_to_index(storage, node_backup)`](https://github.com/thelastpickle/cassandra-medusa/blob/5cb39834e7f8ad1c48641af1f1548898bacd5b51/medusa/index.py#L104) and then [`set_latest_backup_in_index(storage, node_backup)`](https://github.com/thelastpickle/cassandra-medusa/blob/5cb39834e7f8ad1c48641af1f1548898bacd5b51/medusa/index.py#L113
)

Those functions put a record of the backup happening in the storage/index... even though nothing was backed up.
The next line of the test: `Then the backup index exists` would see that something was added to the index and the test continues on happy to delete the (empty) backup that it just made.

Can you confirm that the `differential` backup should return 0 directories?

adejanovski:
Hi @kiddom-kq, is this still something you're working on?

rzvoncek:
ping @kiddom-kq

PR #396, opened 2021-08-24
[WIP] explicitly catch ConnectionRefusedError when determining if a host is up/down

kiddom-kq:
While rolling out Medusa @ my org, I ran into an unhandled exception: `ConnectionRefusedError`

Depending on security-group/firewall/OS configuration, a connection to a closed port may either silently fail (looks like a timeout from the connectors' point of view) or an immediate "no!" in the form of a `RST` packet.

In my case, the host that `medusa` was connecting to was up, but the cassandra process was not and hosts in this network are configured to not waste time with stealthy drops.


adejanovski (approved):
LGTM ✅

PR #397, opened 2021-08-24
[WIP] add support for comments in the hostmap file

kiddom-kq:
While trying to use medusa to restore a snapshot to a cluster, I hit a few issues that broke the cluster. Since the cluster was no longer up, i could not use the `--seed-target` argument. I didn't find the `--host-list` argument documented anywhere so I started taking notes as I looked through the code to figure out what the file was supposed to contain and how it was supposed to be used.

I added rudimentary support for comments in the `host-list` file, too.

I can move the documentation from the function def to somewhere in the docs repo if needed.

### DIFF
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -266,7 +266,7 @@ class RestoreJob(object):
     def _populate_hostmap(self):
         """
         When there are no seed nodes to pull cluster topology from, the essential information required for a restore
-            can be passed in via a simple file using the --host-list CLI argument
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
The `host-list` param and the file structure are documented [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/docs/Restoring-a-full-cluster.md#case-2---restore-to-a-different-cluster-with-same-number-of-nodes), but I reckon it's not necessarily easy to find that information.

Always nice to have more comments in the code and it's cool to have support for comments in the host list file as well.

Thanks!

PR #398, opened 2021-08-24
Update Cassandra test matrix

emerkle826:


emerkle826:
addressed by https://github.com/thelastpickle/cassandra-medusa/pull/401

PR #400, opened 2021-08-31
Add missing deps for Azure-cli in the Docker image

adejanovski:


burmanm (approved):

burmanm:
Careful with the image sizes.

adutra:
Hmm sorry I came too late. I think you should also change the `requirements-azure.txt` file. Currently it has only:

```
azure-cli-core>=2.24.0
azure-mgmt-storage
packaging
```

While it does work as is, it would look more natural if it had just:

```
azure-cli>=2.24.0
```

adutra:
> Careful with the image sizes.

It will get very large for sure. I think the idea of producing different images, one per supported backend, is appealing.

PR #402, opened 2021-09-02
Publish storage provider specific Docker images

adejanovski:
Fixes https://k8ssandra.atlassian.net/browse/K8SSAND-853

Use a matrix to generate provider specific Docker images with a smaller size, while still generating an uber docker image with the dependencies from all providers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,6 +268,9 @@ jobs:
   publish-docker-master:
     needs: [debian-build, build, integration-tests]
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+    strategy:
+      matrix:
+        storage-provider: ["", "-s3", "-azure", "-gcs"]
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


adejanovski:
I did a test CI run with a commit (now removed) which published `tmp-*` tags in Docker hub.
Here's the result:
k8ssandra-medusa-Tags-Docker-Hub (1)

It's working as expected and we have smaller provider specific images as planned.

adutra (approved):

jdonenine (approved):

.github/workflows/ci.yml
```diff
@@ -268,6 +268,9 @@ jobs:
   publish-docker-master:
     needs: [debian-build, build, integration-tests]
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+    strategy:
+      matrix:
+        storage-provider: ["", "-s3", "-azure", "-gcs"]
```
burmanm, line 273: Personal preference, but I would've set "s3", "azure" here and later on to the Dockerfile the ``Dockerfile-${{``

adejanovski: That was my go to as well, but it wouldn't work for the Uber image which has no prefix. Didn't want to get into conditionals here and keep a simple matrix for all images.

.github/workflows/ci.yml
```diff
@@ -277,14 +280,14 @@ jobs:
...
         # Push Docker image tagged with the short commit sha
-        docker build -t k8ssandra/medusa:$(git rev-parse --short HEAD) -f k8s/Dockerfile .
-        docker push k8ssandra/medusa:$(git rev-parse --short HEAD)
+        docker build -t k8ssandra/medusa:$(git rev-parse --short HEAD)${{ matrix.storage-provider }} -f k8s/Dockerfile${{ matrix.storage-provider }} .
+        docker push k8ssandra/medusa:$(git rev-parse --short HEAD)${{ matrix.storage-provider }}
         # Push Docker image tagged as "master"
-        docker tag k8ssandra/medusa:$(git rev-parse --short HEAD) k8ssandra/medusa:master
-        docker push k8ssandra/medusa:master
+        docker tag k8ssandra/medusa:$(git rev-parse --short HEAD)${{ matrix.storage-provider }} k8ssandra/medusa:master${{ matrix.storage-provider }}
+        docker push k8ssandra/medusa:master${{ matrix.storage-provider }}
```
burmanm, line 290: ``:latest`` is more of a common strategy for containers and there's a special handling for them in Kubernetes also. Is there a reason to keep the ``:master`` tag-name instead?

adejanovski: Isn't `:latest` supposed to correspond to the latest release? It's not the case here, we're publishing beta images out of the `master` branch.

PR #403, opened 2021-09-03
Use the aws_session_token setting of the aws profile

atallahade:
### Context

To make medusa authenticate to AWS, I've initially been using environment variables to set the access key, secret key and session/security token ([this part of the code](https://github.com/thelastpickle/cassandra-medusa/blob/18562beaa4d2f4affa6ff143a923b69f748accde/medusa/storage/s3_storage.py#L68-L76)). This worked well when using `medusa` commands, as I would just have to generate the credentials before calling the command, using `aws sts` as [described here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/), because libcloud doesn't automatically assume AMI roles.

I then wanted to use gRPC to issue operations, to be able to use some kind of API and not rely on commands output (especially to list and verify backups). The problem is that I cannot just set the previous environment variables before launching the gRPC server, since the generated credentials eventually expire.

### Description

The changes in this PR will simply allow medusa to set the session/security token, while reading the aws credentials file. The `aws_session_token` setting is coming from [the official doc](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html).
That way, I should be able to generate the credentials periodically and write them to this credentials file. This should work because AFAICS, the gRPC server doesn't cache the credentials but always read the aws credentials file.

### Testing

I couldn't find unit or integration tests covering this part of the code, and since I don't have much experience working with this codebase I thought would just make a PR first. Let me know if you need me to add tests.

I've also tested this change manually on a test Cassandra cluster and this worked well using the credentials file.

### DIFF
--- a/medusa/storage/s3_storage.py
+++ b/medusa/storage/s3_storage.py
@@ -49,7 +49,7 @@ class S3Storage(S3BaseStorage):
 
         :return driver: EC2 driver object
         """
-        aws_security_token = ''
+        aws_session_token = ''
         aws_access_key_id = None
         # or authentication via AWS credentials file
         if self.config.key_file and os.path.exists(os.path.expanduser(self.config.key_file)):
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
LGTM @atallahade, thanks!

I think it would be fairly hard to test this on GHA as it requires to run the test on a VM that has an AMI role, which isn't gonna happen on a GHA runner.

PR #406, opened 2021-09-17
Fix imdsv2 calls from docker

optplx:
This works as is for us, since we currently don't use Docker on our Cassandra nodes. 

But when calling Medusa from Docker, you get a timeout error, since `By default, the response to PUT requests has a response hop limit (time to live) of 1 at the IP protocol level`. Calling from Docker with network bridge for example would give 2 hops (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) .

This can be worked around by using network host, or by increasing the number of hops to 2, but to keep things transparent for Medusa users, it is probably better to catch the timeout exception

WDYT?


adejanovski:
@optplx, the flake8 tests are failing due to a line too long.
Could you fix this so that we can merge the PR?
Thanks!

optplx:
sorry @adejanovski did not notice. I have to say that I am a bit lost with the CI, looks like it is always red somewhere even though it is not always related to my changes

PR #407, opened 2021-09-17
Fix corrupt backups when tables have 2i

adejanovski:
This PR fixes an issues that was reported [here](https://forum.k8ssandra.io/t/medusa-restore-systematically-corruption-of-a-table/328/3).

Secondary indexes are stored as subdirectories of the base table directory. When looking for previously backed up files with differential backups, the file name was the only thing used (per keyspace/table) which leads to some confusion between base table files and 2i files if they have the same name and size.
This leads to corrupt backups with manifests that are missing some files, which are then not restored.

This PR retains the subdirectory for cache comparison for secondary index files, avoiding the confusion between files from the base tables and their 2i.

### DIFF
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -46,7 +46,7 @@ class NodeBackupCache(object):
             self._data_path = node_backup.data_path
             self._cached_objects = {
                 (section['keyspace'], section['columnfamily']): {
-                    pathlib.Path(object['path']).name: object
+                    self._sanitize_file_path(pathlib.Path(object['path'])): object
                     for object in section['objects']
                 }
                 for section in json.loads(node_backup.manifest)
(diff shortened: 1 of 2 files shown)


medusa/backup_node.py
```diff
@@ -65,6 +65,15 @@ def __init__(self, *, node_backup, differential_mode, enable_md5_checks,
...
         self._enable_md5_checks = enable_md5_checks
 
+    def _retain_2i_subdir(self, path):
+        # Secondary indexes are stored as subdirectories to the base table, starting with a dot.
+        # In order to avoid mixing 2i sstables with the base table sstables, the file name isn't enough
+        # to perform the comparison on differential backups. We need to retain the subdir name for 2i tables.
+        if path.parts[-2].startswith('.'):
+            return os.path.join(path.parts[-2], path.parts[-1])
+        else:
+            return path.name
```
jeffbanks, line 75: Checking, is this for backwards compatibility to return the `path.name`?  If it isn't there for some reason due to failure condition, etc. do we need to log that fact?

jeffbanks (dismissed):
lgtm, I left one question regarding the retention of the name if the 2i subdir isn't present.

medusa/backup_node.py
adejanovski: The presence of the dot at the start of the directory name means it's a secondary index and we need to retain the directory name + the file name (`path.parts[-1]` and `path.name` are the same thing).
If that part doesn't start with a dot, it means we're dealing with the base table sstable files, and not the secondary index. In this case, we just need the file name, not the directory it's in.

jeffbanks: 👍🏻

jeffbanks (approved):
Name change looks good!

PR #412, opened 2021-10-06
Set TTL for storage connections

ivanmp91:
This should fix the issues for long running backups using temporary credentials (like AWS IAM role). Right now the problem is those connections are never recycled, and if a backup is running longer than the credentials expiration time, it'll just fail.
This PR set a max TTL for those connections of 20 minutes (it's just an arbitrary number, but should be good enough) and after this period of time, it just resets the connection pool forcing to create new fresh connections.

### DIFF
--- a/medusa/storage/concurrent.py
+++ b/medusa/storage/concurrent.py
@@ -19,6 +19,7 @@ import multiprocessing
 import os
 import pathlib
 import threading
+import time
 
 from libcloud.storage.types import ObjectDoesNotExistError
 from retrying import retry
@@ -39,6 +40,8 @@ class StorageJob:
         self.lock = threading.Lock()
         self.func = func
         self.connection_pool = []
+        self.timer = time.time()
+        self.connection_ttl = 1200
         if max_workers is None:
             self.max_workers = int(multiprocessing.cpu_count())
         else:
@@ -50,6 +53,10 @@ class StorageJob:
 
     def with_storage(self, iterable):
         with self.lock:
+            current_timer = time.time()
+            if current_timer - self.timer > self.connection_ttl:
+                self.connection_pool = []
+                self.timer = time.time()
             if not self.connection_pool:
                 connection = self.storage.connect_storage()
             else:
--- a/medusa/storage/s3_compat_storage/concurrent.py
+++ b/medusa/storage/s3_compat_storage/concurrent.py
@@ -19,6 +19,7 @@ import multiprocessing
 import os
 import pathlib
 import threading
+import time
 
 from libcloud.storage.types import ObjectDoesNotExistError
 from retrying import retry
@@ -41,6 +42,8 @@ class StorageJob:
         self.lock = threading.Lock()
         self.func = func
         self.connection_pool = []
+        self.timer = time.time()
+        self.connection_ttl = 1200
         if max_workers is None:
             self.max_workers = int(multiprocessing.cpu_count())
         else:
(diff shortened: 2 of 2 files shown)


adejanovski:
Hi @ivanmp91,

if I remember our chat correctly, we said that this wouldn't fully fix the issue as the main connection to the storage backend would have the same issue.
Wasn't that correct? Did you conduct tests that allowed you to verify your patch actually fixes the issue in production?

ivanmp91:
> if I remember our chat correctly, we said that this wouldn't fully fix the issue as the main connection to the storage backend would have the same issue. Wasn't that correct?

We talked about two solutions:

- Reimplement the [get_blob](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/s3_compat_storage/awscli.py#L174) method to use aws cli instead. The main problem is that other storage connector would still have the same issue and also it wouldn't get the metadata needed.

- Refresh the credentials keys over the time from the storage connection pool. This is what's implemented in this PR and it fixes the issue where long backups (in the case of relying on AWS IAM role around 6-7 hours) it will start failing because the connection pool is still reusing connections with old keys. Forcing those connections to be refreshed over the time it will prevent to ever use expired credentials and fully fix this issue. Actually in our case the workaround has been to increase the number of threads to make sure the backup is done in less than the credentials expiration time.

>Did you conduct tests that allowed you to verify your patch actually fixes the issue in production?

I haven't tried it in production, but I can do a test with a large data node and make sure the backup takes more than the credentials expiration time to see if it completes successfully. I'll reply back with the results when this is done.

ivanmp91:
@adejanovski sorry it took so long, but finally I've been able to test performing a full backup with the changes introduced in this PR and it worked like a charm at the first attempt without any errors. So it looks like this fixes the issue reported initially in the PR and refreshing the connection pool over time makes the trick:

```
[2022-02-18 22:42:15,858] INFO: Updating backup index
[2022-02-18 22:42:17,965] INFO: Backup done
[2022-02-18 22:42:17,966] INFO: - Started: 2022-02-18 16:48:35
                        - Started extracting data: 2022-02-18 16:49:47
                        - Finished: 2022-02-18 22:42:17
[2022-02-18 22:42:17,966] INFO: - Real duration: 5:52:30.761670 (excludes time waiting for other nodes)
[2022-02-18 22:42:18,054] INFO: - 58376 files, 1.06 TB
[2022-02-18 22:42:18,054] INFO: - 58376 files copied from host
[2022-02-18 22:42:18,139] INFO: Updated from existing status: 0 to new status: 1 for backup id: ivan-test
```

medusa/storage/concurrent.py
```diff
@@ -50,6 +53,10 @@ def execute(self, iterables):
 
     def with_storage(self, iterable):
         with self.lock:
+            current_timer = time.time()
+            if current_timer - self.timer > self.connection_ttl:
+                self.connection_pool = []
+                self.timer = time.time()
```
adejanovski, line 59: Would you mind adding some comments to explain what you're doing here?

medusa/storage/s3_compat_storage/concurrent.py
```diff
@@ -52,6 +55,10 @@ def execute(self, iterables):
 
     def with_storage(self, iterable):
         with self.lock:
+            current_timer = time.time()
+            if current_timer - self.timer > self.connection_ttl:
+                self.connection_pool = []
+                self.timer = time.time()
```
adejanovski, line 61: Same request here, a comment would be nice.

adejanovski (changes requested):
I just have a minor request for comments in the code, and thanks a lot for making those tests after all this time.

PR #418, opened 2021-10-29
Remove unnecessary warning regarding is_ccm

ajmaidak:
Every time a new cassandra is created Medusa prints a WARNING regarding the value of the is_ccm config.

`[2021-10-29 22:00:03,244] WARNING: is ccm : 0`

This appears to have always been the case the blame on this line points to the initial commit from when the code was imported from Spotify...

I can't really see the purpose for this warning so I recommend it be removed.  If it's useful to know the value of ccm this can probably be an INFO.


adejanovski:
You're right, that debug logging has been bugging me for ages but we never took the time to remove it. Thanks!

adejanovski (approved):

PR #423, opened 2021-11-05
Allow GCS tests with 2.2

adejanovski:
Just checking if GCS works with 2.2, please don't review.


adejanovski:
It was just to test GCS with C* 2.2.
Closing.

PR #424, opened 2021-11-08
Update example.ini from docs

elsmorian:
I managed to completely miss the `resolve_ip_addresses = True` option as this is mentioned in the docs but wasn't present in the example ini file.

This PR updates the example ini file with options that are mentioned in the docs but were not present in the `medusa-example.ini` file. https://github.com/thelastpickle/cassandra-medusa/blob/master/docs/Configuration.md


adejanovski (approved):
Thanks for the additions!

PR #430, opened 2021-12-03
Use cassandra native_port for cluster connection

ajmaidak:
If you are using TLS and only listening for CQL TLS connections on port
9142 then medusa will fail to connect because the CqlSessionProvider
only returned a new_session using the default port (9042).

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -56,11 +56,12 @@ class SnapshotPath(object):
 
 class CqlSessionProvider(object):
 
-    def __init__(self, ip_addresses, cassandra_config):
+    def __init__(self, ip_addresses, cassandra_config, port):
         self._ip_addresses = ip_addresses
         self._auth_provider = None
         self._ssl_context = None
         self._cassandra_config = cassandra_config
+        self._native_port = port
 
         if null_if_empty(cassandra_config.cql_username) and null_if_empty(cassandra_config.cql_password):
             auth_provider = PlainTextAuthProvider(username=cassandra_config.cql_username,
@@ -90,6 +91,7 @@ class CqlSessionProvider(object):
          """
 
         cluster = Cluster(contact_points=self._ip_addresses,
+                          port=int(self._native_port),
                           auth_provider=self._auth_provider,
                           execution_profiles=self._execution_profiles,
                           ssl_context=self._ssl_context)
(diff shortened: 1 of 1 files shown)


adejanovski (changes requested):
Hi @ajmaidak,

thanks for pushing this PR!
It seems like you also need to change `verify_restore.py` to add the new port argument to the CqlSessionProvider constructor call in order to fix the integration tests.
There are other source files referencing this constructor that need updating, which aren't covered by integration tests: `backup_cluster.py` and `restore_cluster.py`

ajmaidak:
@adejanovski  thanks for the review, after thinking about this some more and reading the code a bit more.  I'm not sure extending the the signature on the CqlSessionProvider() in this way is ideal.  Unless maybe a a port parameter is added to the [cassandra] section of the ini file?

There seems to have been some attempt to keep the CqlSessionProvider() decoupled from the cassandra.yaml, but in practice only `verify_restore()` doesn't ultimately call the `CassandraConfigReader()`

`restore_cluster()` - Creates a RestoreJob which creates a `Cassandra()` which will call the `CassandraConfigReader()`
`backup_cluster()` - Createsa BackupJob which creates a `Cassandra()` which will call the `CassandraConfigReader()`

It seems like the `verify_restore()` method is attempts to be unaware of the cassandra.yaml.  It receives a config argument but never calls the CassandraConfigReader so changing the code as I'm done creates an implicit dependency on the `verify_restore()` needing access to the cassandra.yaml.  In practice I don't believe this coupling matters as the only way to reach `verify_restore()` is via either a `RestoreJob()` or `restore_node()` both of which will call the CassandraConfigReader long before `verify_restore()` is reached. 

Let me know what you think and if you want this squashed o a single commit.  

If you prefer the signature be extended without adding a config.port parameter I can do so, but likely will need to call the `CassandraConfigReader()` from the `verify_restore()` method before the `CqlSessionProvider()` is initialized.

adejanovski:
@ajmaidak, I agree that it's fine (and probably necessary for consistency) to have `verify_restore()` read the `cassandra.yaml` file just like it's done in other places. Having the native port auto discovered and applying all the rules we've built to apply defaults should be done everywhere we're trying to create a CQL connection.

Definitely -1 on adding `config.port` as we can deal with this in a seamless fashion.

adejanovski (approved):
LGTM, thanks!

PR #431, opened 2021-12-03
Add a datadog statsd monitoring backend

ajmaidak:
This uses the datadog specific statsd package but since only the gauge
metric is used really any statsd package could be used.

The backup-name tag is not sent to datadog to reduce custom metric cost
but can be enabled via the send_backup_name_tag configuration boolean


adejanovski (approved):
Thanks for contributing datadog support!

PR #433, opened 2022-01-14
Fix debian package build with python 3.9

Skunnyk:
- We need a more recent version of setuptools and pip for python 3.9
- Use pybuild buildsystem to prepare the future: https://wiki.debian.org/Python/Pybuild


adejanovski (approved):

adejanovski:
Thanks @Skunnyk !

PR #448, opened 2022-03-08
Support purge operation in the gRPC server

adejanovski:
┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1384) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1384
┆priority: Medium


adejanovski:
I've just rebased to fix the conflicts.
@jeffbanks, do you have spare cycles to review this?

medusa/service/grpc/client.py
```diff
@@ -100,3 +100,13 @@ def backup_exists(self, name):
...
             return False
+
+    def purge_backups(self):
+        try:
+            stub = medusa_pb2_grpc.MedusaStub(self.channel)
+            request = medusa_pb2.PurgeBackupsRequest()
+            resp = stub.PurgeBackups(request)
+            return resp
+        except grpc.RpcError as e:
+            logging.error("Failed to purge backupsdue to error: {}".format(e))
```
jeffbanks, line 111: minor space to include after 'backups'

medusa/service/grpc/medusa.proto
```diff
@@ -71,11 +73,22 @@ message BackupSummary {
...
 }
 
 message BackupNode {
   string         host = 1;
   repeated int64 tokens = 2;
   string         datacenter = 3;
   string         rack = 4;
+}
+
+message PurgeBackupsRequest {
```
jeffbanks, line 86: Do we need to think about supplying any type of identity for the backup request caller?  For example, if we needed to track in the logs the originator of the purge request.  In our case, for now, this would be something like `medusa-operator` or `k8ssandra-operator`.  This could be useful for correlation with logging across systems.

medusa/service/grpc/medusa.proto
```diff
@@ -71,11 +73,22 @@ message BackupSummary {
...
+}
+
+message PurgeBackupsResponse {
+  int32     nbBackupsPurged = 1;
+  int32     nbObjectsPurged = 2;
+  int64     totalPurgedSize = 3;
+  int32     totalObjectsWithinGcGrace = 4;
```
jeffbanks, line 93: Would a `GMT date-time` inclusion be useful for when the purge succeeded for the given totals?

medusa/service/grpc/medusa_pb2_grpc.py
```diff
@@ -74,125 +79,153 @@ def GetBackups(self, request, context):
...
+                    servicer.DeleteBackup,
+                    request_deserializer=medusa__pb2.DeleteBackupRequest.FromString,
+                    response_serializer=medusa__pb2.DeleteBackupResponse.SerializeToString,
+            ),
+            'GetBackups': grpc.unary_unary_rpc_method_handler(
+                    servicer.GetBackups,
+                    request_deserializer=medusa__pb2.GetBackupsRequest.FromString,
+                    response_serializer=medusa__pb2.GetBackupsResponse.SerializeToString,
+            ),
+            'PurgeBackups': grpc.unary_unary_rpc_method_handler(
```
jeffbanks, line 116: Do we have an idea how long this takes to return a response on average?  Would this be something we do `async` and have the controller check back for status?

If so, we would have a corresponding `PurgeStatus` similar to the `AsyncBackup` model.

jeffbanks (changes requested):
Minor nits mostly,  nice work! 

I do have a question about the `sync` vs. `async` approach for the purge operation and how that varies with the options for the backup operations.

medusa/service/grpc/medusa_pb2_grpc.py
adejanovski: Purge should be much faster than backups, so I thought it wasn't necessary for the initial implementation.
Let's get a feel for how that goes for now. Also there is some refactoring planned to make the orchestration code common between MedusaTasks and MedusaBackupJobs which might be a good place to do this.

medusa/service/grpc/medusa.proto
adejanovski: I'm not sure I see how this additional info could be useful 🤔  
The reality is that this call will only be implemented in k8ssandra-operator, so we already know where the calls are coming from.

medusa/service/grpc/client.py
adejanovski: yup, good catch 👍

adejanovski: It's actually fixed in the follow up PR that will be merged into this branch before we merge it to `master`: https://github.com/thelastpickle/cassandra-medusa/pull/449/files#diff-1d56ea287055089d1407ae22c124dfb7e4ad2b53d76c22e04a9c346435ddfc55R112

medusa/service/grpc/medusa.proto
jeffbanks: I'm 🆗 with not adding it in this case.  

It becomes extremely helpful for logging correlation along with date-time completion times when logged.

PR #449, opened 2022-03-08
Allow remote restore in Kubernetes

adejanovski:
This PR changes the way restores are performed in Kubernetes.
They are associated with [this k8ssandra-operator PR](https://github.com/k8ssandra/k8ssandra-operator/pull/454).

Restore orchestration was refactored to allow computing the restore mappings without triggering the restore itself. This restore mapping will be stored under `/var/lib/cassandra/.restore_mapping/<restore key>` when the newly created `PrepareRestore()` gRPC operation is invoked.
Then the medusa-restore container will run `medusa/service/grpc/restore.py` as previously, but it was modified to read the mapping file from the restore preparation phase to find which fqdn it should restore from.
The restore mapping bits were simplified to work both with in place and remote restores (nodes are ordered by smallest token and associated one by one).

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1383) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1383
┆priority: Medium

### DIFF
--- a/medusa/network/hostname_resolver.py
+++ b/medusa/network/hostname_resolver.py
@@ -29,10 +29,9 @@ class HostnameResolver:
             logging.debug("Not resolving {} as requested".format(ip_address_to_resolve))
             return ip_address_to_resolve
 
-        fqdn = socket.getfqdn(ip_address_to_resolve)
-        returned_fqdn = fqdn
-        if self.k8s_mode and fqdn.find('.') > 0:
-            returned_fqdn = fqdn.split('.')[0]
-        logging.debug("Resolved {} to {}".format(ip_address_to_resolve, returned_fqdn))
+        hostname = socket.getfqdn(ip_address_to_resolve)
+        if self.k8s_mode:
+            hostname = socket.getnameinfo((ip_address_to_resolve, 0), socket.NI_NOFQDN)[0]
+        logging.debug("Resolved {} to {}".format(ip_address_to_resolve, hostname))
 
-        return returned_fqdn
+        return hostname
--- a/medusa/network/hostname_resolver.py
+++ b/medusa/network/hostname_resolver.py
@@ -45,7 +45,7 @@ class HostnameResolver:
             reverse_name = dns.reversename.from_address(ip_address).to_text()
             fqdns = dns.resolver.resolve(reverse_name, 'PTR')
             for fqdn in fqdns:
-                if not self.is_ipv4(fqdn.to_text().split('.')[0].replace('-','.')):
+                if not self.is_ipv4(fqdn.to_text().split('.')[0].replace('-', '.')):
                     return fqdn.to_text().split('.')[0]
 
         return ip_address
--- a/medusa/network/hostname_resolver.py
+++ b/medusa/network/hostname_resolver.py
@@ -13,8 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import socket
+import dns.resolver
+import dns.reversename
+import ipaddress
 import logging
+import socket
 
 
 class HostnameResolver:
@@ -23,15 +26,33 @@ class HostnameResolver:
         self.k8s_mode = k8s_mode
 
     def resolve_fqdn(self, ip_address=''):
+        logging.info(f"Resolving ip address {ip_address}")
         ip_address_to_resolve = ip_address if ip_address != '' else socket.gethostbyname(socket.getfqdn())
-
+        logging.info(f"ip address to resolve {ip_address_to_resolve}")
         if str(self.resolve_addresses) == "False":
             logging.debug("Not resolving {} as requested".format(ip_address_to_resolve))
             return ip_address_to_resolve
 
         hostname = socket.getfqdn(ip_address_to_resolve)
         if self.k8s_mode:
-            hostname = socket.getnameinfo((ip_address_to_resolve, 0), socket.NI_NOFQDN)[0]
+            hostname = self.compute_k8s_hostname(ip_address_to_resolve)
         logging.debug("Resolved {} to {}".format(ip_address_to_resolve, hostname))
 
         return hostname
+
+    def compute_k8s_hostname(self, ip_address):
+        if (self.is_ipv4(ip_address)):
+            reverse_name = dns.reversename.from_address(ip_address).to_text()
+            fqdns = dns.resolver.resolve(reverse_name, 'PTR')
+            for fqdn in fqdns:
+                if not self.is_ipv4(fqdn.to_text().split('.')[0].replace('-','.')):
+                    return fqdn.to_text().split('.')[0]
+
+        return ip_address
+
+    def is_ipv4(self, ip_address):
+        try:
+            ipaddress.IPv4Network(ip_address)
+            return True
+        except ValueError:
+            return False
--- a/medusa/network/hostname_resolver.py
+++ b/medusa/network/hostname_resolver.py
@@ -41,11 +41,12 @@ class HostnameResolver:
         return hostname
 
     def compute_k8s_hostname(self, ip_address):
-        if (self.is_ipv4(ip_address)):
+        if (self.is_ipv4(ip_address) or self.is_ipv6(ip_address)):
             reverse_name = dns.reversename.from_address(ip_address).to_text()
             fqdns = dns.resolver.resolve(reverse_name, 'PTR')
             for fqdn in fqdns:
-                if not self.is_ipv4(fqdn.to_text().split('.')[0].replace('-', '.')):
+                if not self.is_ipv4(fqdn.to_text().split('.')[0].replace('-', '.')) \
+                   and not self.is_ipv6(fqdn.to_text().split('.')[0].replace('-', ':')):
                     return fqdn.to_text().split('.')[0]
 
         return ip_address
@@ -53,6 +54,13 @@ class HostnameResolver:
     def is_ipv4(self, ip_address):
         try:
             ipaddress.IPv4Network(ip_address)
+            return True
+        except ValueError:
+            return False
+
+    def is_ipv6(self, ip_address):
+        try:
+            ipaddress.IPv6Network(ip_address)
             return True
         except ValueError:
             return False
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import collections
 import datetime
 import logging
 import operator
@@ -252,7 +251,6 @@ class RestoreJob(object):
         for node in tokenmap.keys():
             nodes[node] = tokenmap[node]['tokens'][0]
         return sorted(nodes.items(), key=operator.itemgetter(1))
-
 
     @staticmethod
     def _is_restore_in_place(backup_tokenmap, target_tokenmap):
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -256,7 +256,10 @@ class RestoreJob(object):
     def _is_restore_in_place(backup_tokenmap, target_tokenmap):
         # If at least one node is part of both tokenmaps, then we're restoring in place
         # Otherwise we're restoring a remote cluster
-        return len(set(backup_tokenmap.keys()) & set(target_tokenmap.keys())) > 0
+        logging.info(f"backup tokenmap keys: {backup_tokenmap.keys()}")
+        logging.info(f"target tokenmap keys: {target_tokenmap.keys()}")
+        logging.info(f"backup tokenmap keys intersection with target tokenmap keys: {set(backup_tokenmap.keys()) & set(target_tokenmap.keys())}")
+        return len(list(set(backup_tokenmap.keys()) & set(target_tokenmap.keys()))) > 0
 
     def _get_seeds_fqdn(self):
         seeds = list()
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -17,6 +17,7 @@
 import collections
 import datetime
 import logging
+import operator
 import socket
 import sys
 import traceback
@@ -129,9 +130,9 @@ class RestoreJob(object):
         self.fqdn_resolver = HostnameResolver(fqdn_resolver, k8s_mode)
         self._version_target = version_target
 
-    def execute(self):
+    def prepare_restore(self):
         logging.info('Ensuring the backup is found and is complete')
-        if not self.cluster_backup.is_complete():
+        if not self.config.kubernetes.enabled and not self.cluster_backup.is_complete():
             raise RuntimeError('Backup is not complete')
 
         # CASE 1 : We're restoring using a seed target. Source/target mapping will be built based on tokenmap.
@@ -149,6 +150,8 @@ class RestoreJob(object):
             self._capture_release_version(session=None)
             logging.info('Starting Restore on all the nodes in this list: {}'.format(self.host_list))
 
+    def execute(self):
+        self.prepare_restore()
         self._restore_data()
 
     @staticmethod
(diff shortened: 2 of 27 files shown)


medusa/restore_cluster.py
```diff
@@ -222,23 +224,13 @@ def _chunk(my_list, nb_chunks):
...
-                backup_ringmap[token].append(host)
-            for token, host in target_tokens.items():
-                target_ringmap[token].append(host)
-
-            self.ringmap = backup_ringmap
-            i = 0
-            for token, hosts in backup_ringmap.items():
-                # take the node that has the same token list or pick the one with the same position in the map.
-                restore_host = target_ringmap.get(token, list(target_ringmap.values())[i])[0]
+            # backup and restore nodes are ordered by smallest token and associated one by one
```
jsanda, line 227: I don't understand why it's necessary to sort the nodes considering target nodes are going to get completely new token assignments.

medusa/restore_cluster.py
```diff
@@ -222,23 +224,13 @@ def _chunk(my_list, nb_chunks):
...
+            sorted_backup_nodes = self._tokenmap_to_sorted_nodes(tokenmap)
+            sorted_target_nodes = self._tokenmap_to_sorted_nodes(target_tokenmap)
+            for i in range(len(sorted_backup_nodes)):
+                restore_host = sorted_target_nodes[i][0]
                 is_seed = True if self.fqdn_resolver.resolve_fqdn(restore_host) in self._get_seeds_fqdn() else False
-                self.host_map[restore_host] = {'source': [hosts[0]], 'seed': is_seed}
-                i += 1
-            logging.debug("self.host_map: {}".format(self.host_map))
+                self.host_map[restore_host] = {'source': [sorted_backup_nodes[i][0]], 'seed': is_seed}
```
jsanda, line 233: Is this safe in this context of k8ssandra? I ask because in the target cluster cass-operator is going to be specifying which nodes are seeds.

medusa/restore_cluster.py
adejanovski: We need a common way of associating nodes, that would give the exact same result when invoked from each pod in the target cluster.
If it's an in place restore, nodes will be associated with themselves (or the node that they replaced). If it's a remote restore, they'll be associated with one of the nodes in a consistent way.
The tokenmaps are not ordered, which made the mapping inconsistent before I did the above change. I would get some pods to map themselves to the same backup node. This is different to what Medusa itself does for restore orchestration as the mapping is done just once when `medusa restore-cluster` is invoked. In the context of k8s, it's invoked once per pod, which requires consistency in the results.

medusa/restore_cluster.py
adejanovski: Actually, and it was a happy surprise, it works perfectly. The intent of specifying which nodes are seeds is to allow Medusa to start the seed nodes first and then the rest of the nodes.
But this is only part of the restore orchestration part which we're not using in the context of k8ssandra. Which is fine, because cass-operator takes care of exactly that: seed nodes are labeled dynamically and nodes start one by one.

jeffbanks (changes requested):
👍🏻 on the clean separation between what is happening with the K8ssandra Medusa operator/controller to ensure existing operations can be leveraged with minor changes.  

I included a couple of suggestions and a question about backup status tracking.

medusa/restore_cluster.py
jsanda: My understanding based on your prior explanations is that the token mappings between  src and dest clusters is done to determine whether it is an in-place or remote restore. In the context of k8ssandra this means that the restore cluster has to be created with all Cassandra pods reaching the ready state and then a restart of all pods to perform the actual restore. 

When used outside of k8s this makes sense since Medusa doesn't really have a true orchestration layer. In k8ssandra-operator though, we do have that orchestration layer. k8ssandra-operator can easily tell Medusa if it is a remote restore. If the operator tells Medusa, is the mapping necessary? If not it would simplify things quite a bit as well as make for a better experience for the user.

medusa/restore_cluster.py
```diff
@@ -126,9 +126,9 @@ def __init__(self, cluster_backup, config, temp_dir, host_list, seed_target, kee
         self.fqdn_resolver = HostnameResolver(fqdn_resolver)
         self._version_target = version_target
 
-    def execute(self):
+    def prepare_restore(self):
         logging.info('Ensuring the backup is found and is complete')
-        if not self.cluster_backup.is_complete():
+        if not self.config.kubernetes.enabled and not self.cluster_backup.is_complete():
```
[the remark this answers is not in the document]
adejanovski: > If k8s is not enabled, should we provide that message so the user knows that the k8s enablement is required?

I don't think so. This condition was introduced because due to hostname resolving issues, backups that are complete may not be seen as such in kubernetes deployments.

Lines 135 to 140 handle both non-remote and remote restores. It's just that you have two ways to trigger a remote restore (in Medusa's orchestration) which is either to provide a seed target and let Medusa do the mapping, or provide a file with the mapping.

medusa/restore_cluster.py
adejanovski: > k8ssandra-operator can easily tell Medusa if it is a remote restore

How so?
k8ssandra-operator would need to check if the topologies match, requiring it to do more work than it does now.

> If the operator tells Medusa, is the mapping necessary? If not it would simplify things quite a bit as well as make for a better experience for the user.

I don't understand what you see that would be simpler than this. The mapping is done automatically by Medusa and the user doesn't need to distinguish in place and remote restores, they're both handled seamlessly as long as you have the same number of nodes.
Medusa needs the mapping to be computed internally anyway for non k8s setups, so this code is necessary. Re-using it seems like a better option than duplicating the logic in the operator.

adejanovski: and we'll soon add more safety around rack assignments, which will be much more easily done by Medusa itself.

medusa/service/grpc/restore.py
```diff
@@ -48,36 +49,55 @@ def configure_console_logging(config):
...
+    keyspaces = {}
+    tables = {}
+    use_sstableloader = False
+
+    cluster_backups = list(medusa.listing.get_backups(config, True))
+    logging.info(f"Found {len(cluster_backups)} backups in the cluster")
+    backup_found = False
+    # Checking if the backup exists for the node we're restoring.
+    # Skipping restore if it doesn't exist.
+    for cluster_backup in cluster_backups:
```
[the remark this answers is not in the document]
adejanovski: is that information stored with the backup in the storage backend? If so, it should be made accessible as a member of the cluster_backup object.
But I think the status you're referring to will only live in memory as long as the server is up. Any backup that was taken before a restart won't have that info.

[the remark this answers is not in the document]
jeffbanks: That is a good point.  I wonder if we should consider having a `backup run` output summary persisted with the storage backend as part of the changes in Medusa.  I don't recall that being something we could or can obtain unless the server is up.

medusa/restore_cluster.py
burmanm: If I understood this correctly, this makes a horrible user experience if everything isn't automated? First the user has to create a K8ssandraCluster and then start the restore process? So the operator might have to be there waiting say an hour for the first process to finish and only then move forward with the operation? To me that feels like an ugly hack and not very Kubernetes-native. That's not what we should be shipping. Lets ignore for a moment the fact how the restore process is done underneath, to user that isn't visible.

What we really should have is a single CRD I create that takes care of everything, including the whole underneath ugly restore processes, starting up correct size Cassandra cluster and giving me the data I originally had. Whether we do this by storing K8ssandraCluster CRD somewhere or calculate the required nodes or make a guess is irrelevant in the first phase. The important thing is that it's a simple process where the operator can simply do one command, go to sleep and come back next morning with an operational cluster. 

Because not only does it reduce the amount of load on the human operator (equals less cost), it also for example allows simple automated testing with the remote restore. I could set my CI/CD job create a staging environment using the latest production data, run the tests and then tear down the setup. And it should all be simple to do. If it requires me to build automation to do the medusa part, then we should do the automation part in to our software. 

There are other benefits in using the operator calculated statistics also. We could be looking at the running Kubernetes cluster and check if there are even resources to load the backup there. Is there enough resources available, enough nodes available, we could do modified Kubernetes scheduling with operator, or we could recalculate new optimal size for cluster (if we could restore to different size - out of this PR, but for future) depending on the Kubernetes cluster size. We could even push for a HPA rule to spin up new nodes in managed Kubernetes cluster to actually get enough resources. Kubernetes is a building block we should leverage and allow us to grow, not limit ourselves to limitations of the old.

If the non-Kubernetes version becomes a burden, we should really consider forking the project to two and not limit our possibilities, just because someone wants to avoid few lines of code.

adejanovski: > this makes a horrible user experience

Well that's not excessive at all 😂 

> What we really should have is a single CRD I create that takes care of everything, including the whole underneath ugly restore processes, starting up correct size Cassandra cluster and giving me the data I originally had. Whether we do this by storing K8ssandraCluster CRD somewhere or calculate the required nodes or make a guess is irrelevant in the first phase. The important thing is that it's a simple process where the operator can simply do one command, go to sleep and come back next morning with an operational cluster.

I'm sure we can build some additional automation on top of what we have in this PR, but I'd like to make that a separate follow up work. There are lots of changes already here and they match the scope that's been designed.

> If the non-Kubernetes version becomes a burden, we should really consider forking the project to two and not limit our possibilities, just because someone wants to avoid few lines of code.

We're not doing it this way out of laziness, there are more reasons to that 🙂 And what you're describing makes that Kubernetes implementation incompatible with backups taken with Medusa on non k8s Cassandra clusters, which I think is a good feature to have.

burmanm: > There are lots of changes already here and they match the scope that's been designed.

I disagree with the definition "scope that's been designed". I don't think we ever agreed on something that's not functional in Kubernetes. Having separate steps which is all due to the fact that this isn't Kubernetes friendly implementation was never the design target. The design specs talked about Helm, which has restrictions on how to create those resources - but those restrictions do not exists with k8ssandra-operator. 

We should rather have a version that doesn't work standalone for restoring than one that doesn't work correctly in Kube. I don't agree with follow-up PR, unless the other PR is ready before this PR is merged. This PR is only few hundred lines, so it's not that big (and some of those lines is generated code).

> We're not doing it this way out of laziness, there are more reasons to that

The previous comment of yours mentioned "duplicate lines" and such, that's not a very good reason in this case when it makes the functionality a lot better. And which it would. 

There's no good UI reason why we wouldn't do this the correct way if we introduce the feature. If there are limitations that prevent this from happening in Kube and non-Kube ends, then we should go with the Kube-only implementation.

> And what you're describing makes that Kubernetes implementation incompatible with backups taken with Medusa on non k8s Cassandra clusters, which I think is a good feature to have.

I didn't say that, the different restore process shouldn't be something that breaks the backup taking capability. A separate future with most work done in Kube-only-Medusa? That could still support older versions (or update path), that shouldn't be an issue either even if it eventually would separate to non-compatible form (eg. use 3.x to remote-restore 3.x backups and 4.x supports only 4.x backups etc). 

The matter of the fact is, we can't have limitations in features just because we need to support legacy. Compromising to both directions is not a good way forward, that has to be cut at some point. "Fork" is just a project management way where we would get rid of the stuff that's not necessary inside Kubernetes and allow us to do stuff that wouldn't work outside without Kubernetes. It would probably even simplify the codebase quite a bit.

medusa/network/hostname_resolver.py
```diff
@@ -23,16 +26,33 @@ def __init__(self, resolve_addresses, k8s_mode):
...
-        logging.debug("Resolved {} to {}".format(ip_address_to_resolve, returned_fqdn))
+        hostname = socket.getfqdn(ip_address_to_resolve)
+        if self.k8s_mode:
+            hostname = self.compute_k8s_hostname(ip_address_to_resolve)
+        logging.debug("Resolved {} to {}".format(ip_address_to_resolve, hostname))
+
+        return hostname
+
+    def compute_k8s_hostname(self, ip_address):
+        if (self.is_ipv4(ip_address)):
```
burmanm, line 44: What happens if the machine has IPv6/IPv4 dual stack? Can the socket.gethostbyname return IPv6 first?

adejanovski: Good call, I'll add support for ipv6 as well.

medusa/service/grpc/restore.py
```diff
@@ -48,36 +49,59 @@ def configure_console_logging(config):
...
-        break
-
-if not backup_found:
-    logging.info("Skipped restore of missing backup {}".format(backup_name))
+if __name__ == '__main__':
+    if len(sys.argv) > 3:
+        config_file_path = sys.argv[2]
+        restore_key = sys.argv[3]
+    else:
+        logging.error("Usage: {} <config_file_path> <restore_key>".format(sys.argv[0]))
```
burmanm, line 57: Where is `sys.argv[1]` ? The helper says 2 parameters, the length check wants 3 (I assume argv[0] is the command).

medusa/restore_cluster.py
```diff
@@ -254,11 +246,19 @@ def _chunk(my_list, nb_chunks):
...
+        return sorted(nodes.items(), key=operator.itemgetter(1))
+
     @staticmethod
     def _is_restore_in_place(backup_tokenmap, target_tokenmap):
         # If at least one node is part of both tokenmaps, then we're restoring in place
         # Otherwise we're restoring a remote cluster
-        return len(set(backup_tokenmap.keys()) & set(target_tokenmap.keys())) > 0
+        logging.info(f"backup tokenmap keys: {backup_tokenmap.keys()}")
+        logging.info(f"target tokenmap keys: {target_tokenmap.keys()}")
+        return len(list(set(backup_tokenmap.keys()) & set(target_tokenmap.keys()))) > 0
```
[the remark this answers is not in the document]
adejanovski: Using the `&` operand allows us to get the entries that exist in both sets. We consider that a restore is an in place one if at least one common node exists in both clusters (some nodes could've been replaced since the backup was taken).
So the expression returns true if the list of overlapping nodes between the two sets is non empty.
I'm not totally sure I need to wrap this in a `list()` though.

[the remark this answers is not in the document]
burmanm: So we need an intersection of set(A) and set(B) ?

Didn't Python have set(A).intersection(set(B)) ? Not sure if you need to cast B even to a set actually. So len of that.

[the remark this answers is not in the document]
burmanm: It seems `set.intersection(set(A), set(B))` would work also as syntax.

medusa/service/grpc/restore.py
```diff
@@ -48,36 +49,59 @@ def configure_console_logging(config):
...
+            #  'host_map':
+            #       {'172.24.0.3': {'source': ['172.24.0.3'], 'seed': False},
+            #        '127.0.0.1': {'source': ['172.24.0.4'], 'seed': False},
+            #        '172.24.0.6': {'source': ['172.24.0.6'], 'seed': False}}}
+            # As each mapping is specific to a Cassandra node, we're looking for the node that maps to 127.0.0.1,
+            # which will be different for each pod.
+            # If hostname resolving is turned on, we're looking for the localhost key instead.
+            if "localhost" in mapping["host_map"].keys():
+                os.environ["POD_IP"] = mapping["host_map"]["localhost"]["source"][0]
+            else:
```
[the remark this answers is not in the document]
adejanovski: ipv6 put aside, it's indeed guaranteed thanks to [this](https://github.com/thelastpickle/cassandra-medusa/pull/449/files/02b0481ab203aeae63b96e0968c9d2e63d49e6b7#diff-0af9023e5dba3232ce866a7482e0b302d746bfa87c8040bb6c3516d481767679R270).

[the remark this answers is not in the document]
adejanovski: I've added a block for ipv6.

burmanm (approved):

Issue #354, opened 2021-05-18
Cassandra backup failed because of mulitple default profiles in ~/.aws/config


giom-l:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13952038)

Since we updated our production environment with 

* cassandra 3.11.9 
* medusa 0.10.0 (was 0.7.1) 

we ran into some strange issues that often prevent backup to end up well.
By often, I mean that some time it runs well, and sometime if fails (and not always the same nodes)

For an unknown reason, some nodes fail with the following error : 

```
[1] 12:00:24 [FAILURE] <ip> Exited with error code 1
Stderr: [2021-05-18 12:00:23,717] DEBUG: Loading configuration from /etc/medusa/medusa.ini
[2021-05-18 12:00:23,719] DEBUG: Logging to file options: LoggingConfig(enabled='1', file='/var/log/medusa/medusa.log', format='[%(asctime)s] %(levelname)s: %(message)s', level='INFO', maxBytes='2000000', backupCount='2')
[2021-05-18 12:00:23,720] INFO: Monitoring provider is noop
[2021-05-18 12:00:23,720] DEBUG: Loading storage_provider: s3_eu_west
[2021-05-18 12:00:23,720] DEBUG: Getting IAM Role:
[2021-05-18 12:00:23,726] DEBUG: Reading AWS credentials from IAM Role: cassandra12-20200313151108838000000004
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/botocore/configloader.py", line 149, in raw_config_parse
    cp.read([path])
  File "/usr/lib64/python3.7/configparser.py", line 696, in read
    self._read(fp, filename)
  File "/usr/lib64/python3.7/configparser.py", line 1065, in _read
    lineno)
configparser.DuplicateSectionError: While reading from '/home/cassandra/.aws/config' [line  4]: section 'default' already exists

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/local/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/usr/local/lib/python3.7/site-packages/awscli/clidriver.py", line 69, in main
    driver = create_clidriver()
  File "/usr/local/lib/python3.7/site-packages/awscli/clidriver.py", line 78, in create_clidriver
[...]

PR #453, opened 2022-03-24
refactor: invoke CLIs as Python modules

maxbrunet:
Just an improvement suggestion, I think it would be more robust to let Python find the CLIs via its `PYTHONPATH`.

Basically, this does `python -m awscli` and `python -m azure.cli` with the right Python executable for `awscli` and `azure-cli` respectively.

I think the `aws_cli_path` setting should be removed or at least deprecated. Let me know what you think.

Later, it could be good to consider replacing the `awscli` by [s3transfer](https://github.com/boto/s3transfer) (the library it uses under the hood), it would more lightweight and would avoid odd issues, especially with the `max_bandwidth` config (#354), but it is not GA yet.

### DIFF
--- a/medusa/storage/azure_blobs_storage/azcli.py
+++ b/medusa/storage/azure_blobs_storage/azcli.py
@@ -12,6 +12,7 @@ from medusa.storage.abstract_storage import AbstractStorage
 
 
 class AzCli(object):
+
     def __init__(self, storage):
         self._config = storage.config
         self.storage = storage
@@ -35,7 +36,7 @@ class AzCli(object):
                 AZURE_STORAGE_ACCOUNT=credentials['storage_account'],
                 AZURE_STORAGE_KEY=credentials['key']
             )
-        self._az_cli_path = self.find_az_cli()
+        self._az_cli_cmd = self.cmd()
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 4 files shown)


adejanovski:
Hi @maxbrunet,

I love the idea of doing these simplifications. It looks like it broke the Azure tests though: https://github.com/thelastpickle/cassandra-medusa/runs/5674058431?check_suite_focus=true#step:8:224

I'm all for ditching the awscli if we can find a library that's as efficient and allows us not to run commands, while retaining compatibility with all S3 clones. Maybe this should go upstream in libcloud though, don't you think?

maxbrunet:
>  It looks like it broke the Azure tests though

Oh I had run the CI before opening the PR, but I had not noticed Azure was skipped because I do not have Azure credentials. It should work if the `azure-cli` package is installed:

```shell
python -c 'import subprocess, sys; subprocess.check_call([sys.executable, "-m", "azure.cli", "help"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)'
```

I cannot find where the `AZURE` dependency group is installed from, could tests have been using a preinstalled CLI on the runner before which cannot be imported?
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#cli-tools

> I'm all for ditching the awscli if we can find a library that's as efficient and allows us not to run commands, while retaining compatibility with all S3 clones. Maybe this should go upstream in libcloud though, don't you think?

`libcloud` seems to only depend on `requests`, so I suppose they aim to keep it that way

maxbrunet:
> Got to dig again into Azure stuff to see what's failing here.

a8a61b88a0ff204cf39cd8c08ee1662bf2421a6a should fix it

adejanovski:
> > Got to dig again into Azure stuff to see what's failing here.
> 
> [a8a61b8](https://github.com/thelastpickle/cassandra-medusa/commit/a8a61b88a0ff204cf39cd8c08ee1662bf2421a6a) should fix it

I figured out what was wrong, finally. The Azure cli has a bug in 2.34 that prevents the use of a flag we're using, `--content-md5` (see [this issue](https://github.com/Azure/azure-cli/issues/21494)).

Also, the azure-cli was pre-installed on Ubuntu in the images we're using in CI. I've had to uninstall it first and change the azure requirements to install 2.33.1 at most, and then the tests passed.

adejanovski (approved):
Since issues in Azure integration tests are unrelated to these changes, we can move forward and merge the PR. Thanks @maxbrunet

maxbrunet:
> The Azure cli has a bug in 2.34 that prevents the use of a flag we're using, `--content-md5` (see [this issue](https://github.com/Azure/azure-cli/issues/21494)).

I wonder if dependencies should be better locked, migrating from `setuptools` to [Poetry](https://python-poetry.org/) which has a lockfile could be a solution, and [Renovate](github.com/renovatebot/renovate) could be used to automate updates. I could help if there is interest?

PR #456, opened 2022-03-25
chore: remove GCS dependency group

maxbrunet:
I have looked around and I cannot find where this package is used. `libcloud` appears to do everything. It is not even installed by the CI or [Dockerfile-gcs](https://github.com/thelastpickle/cassandra-medusa/blob/master/k8s/Dockerfile-gcs). Please do not hesitate to let me know if I am missing something

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1377) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1377
┆priority: Medium

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -17,7 +17,7 @@ Install Medusa on each Cassandra node using one of the following methods.
 
 * if the storage backend is a locally accessible shared storage, run `sudo pip3 install cassandra-medusa`
 * if your backups are to be stored in AWS S3 or S3 compatible backends (IBM, OVHCloud, MinIO, ...), run `sudo pip3 install cassandra-medusa[S3]`
-* if your backups are to be stored in Google Cloud Storage, run `sudo pip3 install cassandra-medusa[GCS]`
+* if your backups are to be stored in Google Cloud Storage, run `sudo pip3 install cassandra-medusa`
 * if your backups are to be stored in Azure Blob Storage, run `sudo pip3 install cassandra-medusa[AZURE]`
 
 Running the installation using `sudo` is necessary to have the `/usr/local/bin/medusa` script created properly.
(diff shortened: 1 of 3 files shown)


adejanovski:
Actually, it should be required to get access to GSutil, which is used for sstables uploads/downloads.

maxbrunet:
hmm `google-cloud-storage` does not provide `gsutil`, this package would: https://pypi.org/project/gsutil/

`gsutil` must be installed by this step in the image:

https://github.com/thelastpickle/cassandra-medusa/blob/3c01bef73bf266f84c6c7357b07c72a8a5386b86/k8s/Dockerfile-gcs#L59-L67

```shell
$ docker run --rm --entrypoint= docker.io/k8ssandra/medusa:0.12.1-gcs which gsutil
/home/cassandra/google-cloud-sdk/bin/gsutil
```

And the CI must use the preinstalled Google Cloud SDK: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md#cli-tools

maxbrunet:
@adejanovski If PR #453 looks good to you, it can be merged it. And I can propose a separate rework of `gsutil` / GCS dependencies in the same fashion as S3/Azure, so it is easier to review. Then this PR can be declined.

adejanovski:
> @adejanovski If PR https://github.com/thelastpickle/cassandra-medusa/pull/453 looks good to you, it can be merged. And I can propose a separate rework of gsutil / GCS dependencies in the same fashion as S3/Azure, so it is easier to review. Then this PR can be declined.

Sounds good. I've merged #453 and would happily accept a PR reworking the GCS dependencies the same way 👍 
So you'd close this PR first and create a new one, right?

maxbrunet:
> So you'd close this PR first and create a new one, right?

I will keep this one open as a simple tracking issue until the other PR is merge if that's ok

maxbrunet:
Invoking `gsutil` as a Python module would be `python -m gslib`, but looking at it again there is a dependency on the `gcloud` CLI:
https://github.com/thelastpickle/cassandra-medusa/blob/757ce53c6943930706d10244a62e798b85e504b6/medusa/storage/google_cloud_storage/gsutil.py#L39-L40
I am not familiar enough with Google Cloud to tell if we could have an alternative without the whole SDK bundle.

And this step in documentation would not be valid anymore as the bundle comes with its own isolated Python environment, so `python -m gslib` cannot work within the Medusa runtime:
https://github.com/thelastpickle/cassandra-medusa/blob/757ce53c6943930706d10244a62e798b85e504b6/docs/Installation.md?plain=1#L72

Adding an explicit Python dependency on `gsutil` would redundant imo

If you are fine to continue using the SDK bundle, this PR can be merged as-is since `google-cloud-storage` is truly unused

adejanovski (approved):
Sure thing, let's move on then.

PR #458, opened 2022-03-30
ISSUE-457: Remove plaintext passwords in config

ossarga:
* Extended functionality for CQLSH credentials to allow nodetool credentials and keystore/truststore passwords to be defined in environment variables
* Updated environment variables to be prefixed with `MEDUSA_`
* Added deprecated warning message to `CQL_*` environment variables
* Minor change to tox configuration to allow individual unit tests to be executed

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1375) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1375
┆priority: Medium

### DIFF
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -211,10 +211,27 @@ def parse_config(args, config_file):
         # Use the ip address instead of the fqdn when DNS resolving is turned off
         config['storage']['fqdn'] = socket.gethostbyname(socket.getfqdn())
 
-    if "CQL_USERNAME" in os.environ:
-        config['cassandra']['cql_username'] = os.environ["CQL_USERNAME"]
-    if "CQL_PASSWORD" in os.environ:
-        config['cassandra']['cql_password'] = os.environ["CQL_PASSWORD"]
+    for config_property in ['cql_username', 'cql_password']:
+        config_property_upper_old = config_property.upper()
+        config_property_upper_new = "MEDUSA_{}".format(config_property.upper())
+        if config_property_upper_old in os.environ:
+            config['cassandra'][config_property] = os.environ[config_property_upper_old]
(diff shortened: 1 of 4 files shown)


medusa/config.py
```diff
@@ -211,10 +211,17 @@ def parse_config(args, config_file):
...
+        'cql_username',
+        'cql_password',
+        'nodetool_username',
+        'nodetool_password',
+        'sstableloader_tspw',
+        'sstableloader_kspw'
+    ]:
+        config_property_upper = "MEDUSA_{}".format(config_property.upper())
+        if config_property_upper in os.environ:
+            config['cassandra'][config_property] = os.environ[config_property_upper]
```
adejanovski, line 224: Would it be possible to still support `CQL_USERNAME` and `CQL_PASSWORD`, while adding a deprecation message if they're being used?

I'm not totally comfortable with such a breaking change.

adejanovski (changes requested):
I'm ok with the PR, at the exception of the immediate removal of CQL_USERNAME and CQL_PASSWORD env variables.
We should deprecate them with a proper message and remove them in the future.

medusa/config.py
ossarga: You have raised a good point. I agree; I think supporting both and adding a deprecation message is a good idea.

I will implement this as well.

ossarga:
@adejanovski thank you for the initial review of the PR. I've updated it as suggested. Let me know what you think about the deprecation message.

adejanovski (approved):
LGTM, thanks @ossarga !

ossarga:
Thank you @adejanovski for the second review and merging the patch!

PR #459, opened 2022-03-31
Add missing documentation for configuration options

Dutchy-:
Add documentation for the option `nodetool_ssl` as used here:

https://github.com/thelastpickle/cassandra-medusa/blob/a00a45593c44b3e4f99adaa1a89475af987c1a9b/medusa/nodetool.py#L21

Note that it may be worth it to change it to a proper boolean instead.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -15,6 +15,17 @@ Modify it to match your requirements:
 ;nodetool_password_file_path = <path to nodetool password file>
 ;nodetool_host = <host name or IP to use for nodetool>
 ;nodetool_port = <port number to use for nodetool>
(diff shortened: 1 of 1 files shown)


Dutchy-:
I only just now noticed that the option is already available in 
https://github.com/thelastpickle/cassandra-medusa/blob/93b46e8073602751bd74c4854edab800c605960a/medusa-example.ini

Dutchy-:
I've updated the PR with all options that exist in medusa-example.ini but not in the documentation.

adejanovski (approved):

adejanovski:
Thanks for fixing the docs!

Issue #468, opened 2022-04-23
Medusa currently doesn't support Python3.10 (Ubuntu 22.04)


andyfoston:
When attempting to run medusa on Python 3.10 on Ubuntu 22.04, I get the following error:

```bash
$ medusa
Traceback (most recent call last):
  File "/usr/local/bin/medusa", line 5, in <module>
    from medusa.medusacli import cli
  File "/usr/local/lib/python3.10/dist-packages/medusa/medusacli.py", line 40, in <module>
    import medusa.backup_cluster
  File "/usr/local/lib/python3.10/dist-packages/medusa/backup_cluster.py", line 24, in <module>
    from medusa.orchestration import Orchestration
  File "/usr/local/lib/python3.10/dist-packages/medusa/orchestration.py", line 18, in <module>
    from pssh.clients.ssh import ParallelSSHClient
  File "/usr/local/lib/python3.10/dist-packages/pssh/clients/__init__.py", line 19, in <module>
    from .native.parallel import ParallelSSHClient
  File "/usr/local/lib/python3.10/dist-packages/pssh/clients/native/__init__.py", line 19, in <module>
    from .parallel import ParallelSSHClient
  File "/usr/local/lib/python3.10/dist-packages/pssh/clients/native/parallel.py", line 20, in <module>
    from .single import SSHClient
  File "/usr/local/lib/python3.10/dist-packages/pssh/clients/native/single.py", line 29, in <module>
    from ssh2.session import Session, LIBSSH2_SESSION_BLOCK_INBOUND, LIBSSH2_SESSION_BLOCK_OUTBOUND
  File "ssh2/session.pyx", line 1, in init ssh2.session
  File "ssh2/channel.pyx", line 1, in init ssh2.channel
  File "ssh2/sftp.pyx", line 1, in init ssh2.sftp
ImportError: /usr/local/lib/python3.10/dist-packages/ssh2/sftp_handle.cpython-310-x86_64-linux-gnu.so: undefined symbol: _PyGen_Send
```

It seems that the `python-ssh2` library added python 3.10 support in their release 0.27.0 https://github.com/ParallelSSH/ssh2-python/releases/tag/0.27.0, and Medusa has their dependency pinned to 0.22.0. Please can this pinning be changed to 0.27.0 to allow Medusa to run on Python 3.10?

[...]

PR #469, opened 2022-04-23
Support python 3.10

andyfoston:
Add support for Python 3.10. Required for supporting Ubuntu 22.04

Ticket: https://github.com/thelastpickle/cassandra-medusa/issues/468

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1472) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1472
┆priority: Medium


.github/workflows/ci.yml
```diff
@@ -104,7 +106,7 @@ jobs:
         fail-fast: false
         matrix:
           #python-version: [3.6]
-          python-version: [3.6, 3.9]
+          python-version: [3.6, 3.9, "3.10"]
```
adejanovski, line 109: We can drop `3.9` here. Testing for `3.6` and `3.10` should hopefully be enough, since we already have a build test for each supported python version.

adejanovski (changes requested):
The latest JDK 8/11 releases introduce a change that break nodetool.
You'll need to make [this change](https://github.com/thelastpickle/cassandra-medusa/pull/449/commits/57c64e4d18113060855db5e6508f99bbcc6fcca9) so that CI can pass.
Also it's not necessary to run the 3.9 integration tests if we cover 3.10.

.github/workflows/ci.yml
```diff
@@ -125,25 +127,25 @@ jobs:
             - it-backend: s3
               cassandra-version: 'github:apache/trunk'
             - it-backend: s3
-              python-version: 3.9
+              python-version: "3.10"
```
andyfoston, line 130: I've changed these excludes from 3.9 to 3.10 now that 3.9 integration tests are no longer being run. Let me know if this was wrong, and I'll revert this part.

adejanovski: That's exactly what I had in mind 👍

adejanovski:
It looks like we have several problems here. The debian build phase is failing, and I remember having to pin setuptools to the version we have in `master`, otherwise it wouldn't build anymore :(
Also, it looks like the Cassandra python driver we're using could be incompatible with python 3.10 (unless ccm is the problem?).

andyfoston:
> It looks like we have several problems here. The debian build phase is failing, and I remember having to pin setuptools to the version we have in `master`, otherwise it wouldn't build anymore :( Also, it looks like the Cassandra python driver we're using could be incompatible with python 3.10 (unless ccm is the problem?).

For the integration tests, it appears that the SSL cipher is no longer permitted in python3.10's SSL library. In fact, all of the ciphers in `ssl_context.get_ciphers()` are for >= TLSv1.2. I'm going to update the tests to support a different cipher/min TLS version for python 3.10 only. 

For the debian build. I had to change the pinning of setuptools to get this to work locally, but for some reason this doesn't work in github actions. I'll look into this further

tests/integration/features/steps/integration_steps.py
```diff
@@ -115,10 +116,16 @@ def cleanup_storage(context, storage_provider):
 
 
 def get_client_encryption_opts(keystore_path, trustore_path):
+    # Python 3.10 has a reduced set of ciphers available. Ensure that a compatible cipher is used for Py310, and
+    # allow older python versions to test older ciphers/TLS versions.
+    if sys.version_info >= (3, 10):
```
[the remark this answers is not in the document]
adejanovski: it looks good to me 👍

adejanovski:
@andyfoston, the PR looks like it's ready to get merged, but you'll need to rebase first and fix the conflicts that should be related to the 0.13 release.
We also have a fix now in `master` for the jdk upgrade issues. The build will now work without needing to pin a specific jdk version.

andyfoston:
> @andyfoston, the PR looks like it's ready to get merged, but you'll need to rebase first and fix the conflicts that should be related to the 0.13 release. We also have a fix now in `master` for the jdk upgrade issues. The build will now work without needing to pin a specific jdk version.

Ah, great. I've updated this now and have reverted my upgrade of setuptools as this didn't have the intended effect (this issue is fixed in https://github.com/thelastpickle/cassandra-medusa/pull/475 anyway)

adejanovski (approved):
Awesome stuff @andyfoston, thanks a lot for your contributions!

PR #475, opened 2022-05-12
Fix debian builds

andyfoston:
Include missing build dependencies
Ensure setuptools isn't upgraded during pip install step

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1507) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1507
┆priority: Medium


adejanovski:
Hi @andyfoston, should that PR fix the failing Focal builds?
If so, could you re-enable those in CI so we can see if it works?
The matrix needs to be updated [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/.github/workflows/ci.yml#L62-L66).

Thanks!

andyfoston:
> Hi @andyfoston, should that PR fix the failing Focal builds? If so, could you re-enable those in CI so we can see if it works? The matrix needs to be updated [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/.github/workflows/ci.yml#L62-L66).
> 
> Thanks!

Hi, yeah this does. Sorry, I didn't spot that this had been commented out. I'll enable it

.github/workflows/ci.yml
```diff
@@ -308,11 +308,10 @@ jobs:
     strategy:
       matrix:
         # Focal builds are broken in CI, we can only build bionic for now
-        #suite: [focal, bionic]
-        suite: [bionic]
+        suite: [focal, bionic]
```
adejanovski, line 311: This is the publish job for the debian packages, which only happens at release time. We have another job at the top of the workflow that builds the debian packages and tests that they install fine.

andyfoston-sky: Ah, sorry, my mistake. I've fixed this now

adejanovski:
It works!!!

Thanks @andyfoston, you're a life savior!! 🙏 🥳

PR #487, opened 2022-06-24
Enable overwriting blobs on retries and remove md5 hash computation on azcli uploads

### LINKED ISSUE #489: Excessive memory consumption with Azure storage backend
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13937863)

Memory consumption of Medusa with the Azure storage backend can grow excessively to 10s of GB.
The behavior was noted to be more important when the azcli is used over libcloud.

It looks like the `--content-md5` argument was used for each azcli invocation, with a computation of the hash made by Medusa through a static method. It's safe to infer that computing the hash for each uploaded file can cause such memory consumption, which could leak and not be garbage collected.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1600) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1600
┆priority: Medium

adejanovski:
Fixes #489 

It's been reported [on Discord](https://discord.com/channels/836217371453685760/836217371453685763/988451041622241300) that the Medusa container in K8ssandra could consume huge amounts of memory when using the azcli for uploads.
It looks like the Azure storage code was setting the `--content-md5` argument with the md5 hash of the uploaded file, which was computed through a static method.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1598) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1598
┆priority: Medium


medusa/storage/azure_blobs_storage/azcli.py
```diff
@@ -54,7 +52,7 @@ def cp_upload(self, *, srcs, bucket_name, dest, max_retries=5):
         # Az cli expects the client to provide the MD5 hash of the upload
         for src in srcs:
             cmd = self._az_cli_cmd + ["storage", "blob", "upload", "-f", str(src), "-c", bucket_name, "-n", dest,
-                                      "--content-md5", AbstractStorage.generate_md5_hash(src)]
+                                      "--overwrite", "--validate-content"]
```
jsanda, line 55: Does `--validate-content` cause the md5 checksum to get computed for each chunk? If so, isn't that what we're trying to avoid?

adejanovski: I can remove this as it's additional cpu/memory load, but since we're invoking the azcli using a subprocess, memory should be freed as soon as the upload finishes 😕 
I'll amend the PR to remove it nonetheless, md5 checks can be enabled in config anyway.

jsanda (approved):

Issue #338, opened 2021-04-22
How to enable AWS encryption with KMS ?


mgrenonville:
Hello, 

Thanks for this project !
I'm trying to enable encryption on S3 bucket, and it looks like it's possible since I've found this issue from jfharden on libcloud project https://github.com/apache/libcloud/issues/1401

Is there a mini-how to to help me in this task ?

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-197) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-197

adejanovski:
While libcloud allows this, we don't seem to support it in Medusa.
We already have a [similar flag](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa-example.ini#L126) to disable md5 checks, which could be used to set `verify_hash=False` on S3 uploads [here](https://github.com/thelastpickle/cassandra-medusa/blob/3c01bef73bf266f84c6c7357b07c72a8a5386b86/medusa/storage/s3_compat_storage/concurrent.py#L133-L135).

We happily accept PRs ;)

justinmir:
I've taken a stab at an implementation for KMS support in Medusa. I can iterate on the implementation or look at alternative approaches if another is more ergonomic.

PR #506, opened 2022-07-18
Support for specifying KMS key for encrypting backups server-side on S3.

justinmir:
This PR enables server-side encryption for backups stored in S3 via S3-KMS. A kms key with `kms_id`
can be specified in the configuration.

To accomplish this the PR does the following:

* `kms_id` configuration parameter for use with the s3 storage provider.
* Support for each storage provider to pass headers to `libcloud` to enable specifying KMS key parameters for S3. 
* s3_compact_storage may pass kms command line parameters to aws s3 cp.

## Testing
Tested with an integration test scenario 1 with a configuration that includes `kms_id` and a bucket configured
with the associated `kms` key.

Fixes: #338

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1674) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1674
┆priority: Medium


adejanovski (changes requested):
Hi @justinmir,

first of all, thanks a lot for pushing this PR!

Since data is encrypted in the bucket, I assume libcloud needs to be able to read them decrypted, right? Shouldn't we then pass the headers in the read blob method [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/abstract_storage.py#L105-L107)?

We're going to need an integration test for that feature, similar to the one you've tested.
I guess the simplest thing we could do is use the default KMS key, which apparently gets used when `--sse aws:kms` is used in the aws cli command but no key is specified.
That will require some changes as you need to allow configuring KMS without specifying a key id. Two options here: create another configuration setting to enable/disable KMS encryption or use a special value such as `default` for the kms_id which would be handled appropriately.
Both the libcloud headers and the awscli commands would need to be adjusted to handle both cases.

justinmir:
Appreciate you taking a look at this!

> Since data is encrypted in the bucket, I assume libcloud needs to be able to read them decrypted, right? Shouldn't we then pass the headers in the read blob method [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/abstract_storage.py#L105-L107)?

The client does not specify the `amz-server-side-encryption` headers in the GET request path for KMS encrypted objects. The KMS key is stored with the object, and AWS checks that the client has permissions to decrypt the KMS key.

> We're going to need an integration test for that feature, similar to the one you've tested.
I've added an integration test for the feature that uses the default S3 KMS key: `alias/aws/s3`, this is the key that is used if `kms_id` is not passed. It currently uses the `medusa-it-gha` bucket, however, it may be better to setup a separate bucket that denies unencrypted requests (since the integration test should pass even if my code doesn't do encryption). I've verified this passes on my own bucket with that setup.

Happy to implement either of the suggested approaches if that is preferred over using `alias/aws/s3`: support for `default` or adding another configuration parameter such as `kms_encryption: true`, where `kms_id` may be set to specify the key.

**Sources**
https://aws.amazon.com/premiumsupport/knowledge-center/decrypt-kms-encrypted-objects-s3/
> You don’t need to specify the AWS KMS key ID when you download an SSE-KMS-encrypted object from an S3 bucket. Instead, you need the permission to decrypt the AWS KMS key.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-kms-encryption.html
> If your object uses SSE-KMS, don't send encryption request headers for GET requests and HEAD requests
> The response headers of the following REST API operations (GET, HEAD, etc.) return the x-amz-server-side-encryption header when an object is stored using server-side encryption.

justinmir:
Hi @adejanovski, wanted to follow-up to see if I can get some :eyes: on this.

adejanovski:
Hi @justinmir,

I've just pushed a commit to your branch, using a new S3 bucket that I created with a custom KMS key. I hope that's ok with you.
Let's see how integration tests run.

adejanovski:
Actually my push to your fork didn't work. I ran the new test locally with success, and pushed another PR to get CI passing with the S3 tests.
Once it passes, you can cherry pick dd43dd8c573b13224ff759d2756a22ea8474f873 and we'll be good for a merge.

adejanovski:
@justinmir, I got the S3 tests to pass with the encrypted bucket. The commit you'll need to cherry pick is d15db1ea42af9aebf0486dc33933d16b25de5e24.

Here's a successful run: https://github.com/thelastpickle/cassandra-medusa/actions/runs/3181032142/jobs/5185339402

justinmir:
Awesome :), I've cherry-picked the commit.

adejanovski (approved):
Awesome stuff! thanks a lot @justinmir

PR #521, opened 2022-09-13
Rack aware cluster restore

anclrii:
Adding `--match-racks` restore option that enables rack aware host matching. This is necessary when using rack aware snitch configuration with nodes in multiple racks. (Without this replicas may be restored into the same rack on the target cluster and restored data will have missing replicas.)

* `_validate_ringmap` checks if target cluster topology matches backup cluster
* `host_map` is populated based on rack placement of the nodes.
* Racks are matched based on node count in backup and target racks, rack names are not considered.
* Nodes are matched so that all data backed up from any one rack is restored to a target rack of the same size.
* The change is backward compatible when `--match-racks` option is not set.
* Added some unit tests.


anclrii:
Related to https://github.com/thelastpickle/cassandra-medusa/issues/365

medusa/restore_cluster.py
```diff
@@ -225,14 +243,29 @@ def _chunk(my_list, nb_chunks):
...
             for i in range(len(sorted_backup_nodes)):
                 restore_host = sorted_target_nodes[i][0]
                 is_seed = True if self.fqdn_resolver.resolve_fqdn(restore_host) in self._get_seeds_fqdn() else False
                 self.host_map[restore_host] = {'source': [sorted_backup_nodes[i][0]], 'seed': is_seed}
+
+        elif topology_matches and self.match_racks:
+            # restore data from the same backup rack into a single rack with matching node count
+            backup_nodes_per_rack = self._tokenmap_to_nodes_per_rack(tokenmap)
+            target_nodes_per_rack = self._tokenmap_to_nodes_per_rack(target_tokenmap)
+            for i in sorted(backup_nodes_per_rack, key=lambda k: len(backup_nodes_per_rack[k]), reverse=True):
```
adejanovski, line 259: There could be a problem here. NetworkTopologyStrategy will place the next replica on the next node in token order from the next rack. The next rack is picked on alphabetical order, which is not considered here.
That's not a problem as long as nb_racks == rf, but as soon as we have a number of racks different than RF, it will start being a problem.

The way I see it, we should check that after sorting both backup and target racks in alphabetical order, they have the same number of nodes. Associating racks should then be done by that alphabetical order only.
If corresponding racks don't have the same number of nodes, then it means the topologies don't match and we need to fallback to sstableloader.

wdyt?

adejanovski (commented):
Hi @anclrii,

it's been a long time since I was hoping for such a contribution, thanks a lot!
I have some concerns about the way racks are ordered in your code (by node count) vs how NetworkTopologyStrategy orders rack to place replicas.
Let me know what you think of my comments.

medusa/restore_cluster.py
anclrii: I did not consider this case at all, that's a good point. Also I didn't know about the alphabetical order. I'll update the logic to what you described.

adejanovski (changes requested):
LGTM, there's just one method that can be simplified since the nodes don't need to be sorted to perform the mapping.

Aside from that, we're good to go 👍

medusa/restore_cluster.py
```diff
@@ -252,6 +281,18 @@ def _tokenmap_to_sorted_nodes(self, tokenmap):
...
+    def _tokenmap_to_nodes_per_rack(self, tokenmap):
+        # Returns an alphabetically sorted list of racks
+        nodes_per_rack = dict()
+        for node in tokenmap.keys():
+            rack = tokenmap[node].get('rack', "")
+            token_0 = tokenmap[node].get('tokens', [""])[0]
+            nodes_per_rack.setdefault(rack, []).append((node, token_0))
+        for rack, nodes in nodes_per_rack.items():
+            sorted_nodes = sorted(nodes, key=operator.itemgetter(1))
+            nodes_per_rack[rack] = sorted_nodes
```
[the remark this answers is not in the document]
anclrii: Makes sense, that's removed now.

adejanovski (approved):
Looks great! Thanks a lot for pushing this PR.

mohammad-aburadeh:
Hi @adejanovski @anclrii 

Any idea when this feature will be released? 
We have more than 25 production clusters that are multi-reacks. We will not be able to restore the backups till we have this feature. 

It's really urgent to have this feature soon. 
Is there any workaround solution to restore the backup on multi-rack clusters? 

Regards
Mohammad

adejanovski:
Hi @mohammad-aburadeh, sure thing, it's been a while now and I see no reason not to cut out a new release.
I'll try to do this within a week.

mohammad-aburadeh:
Thanks @adejanovski

PR #536, opened 2022-10-17
enable sourcing environment file within wrapper

rtib:
When using medusa with an own-hosted S3 storage available via https an environment variable `SSL_CERT_FILE` needs to be set up. The backup-cluster subcommand, however, would fail, as the environment variables aren't sent via SSH. This patch allows the medusa-wrapper to source an environment file before calling the medusa backup on the local nodes.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1839) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1839
┆priority: Medium

### DIFF
--- a/bin/medusa-wrapper
+++ b/bin/medusa-wrapper
@@ -20,6 +20,9 @@ if test -f pid; then
     done
     exit $(cat status)
 fi
+if [ -f /etc/default/cassandra-medusa ]; then
+    . /etc/default/cassandra-medusa
+fi
 $@ >stdout 2>stderr &
 echo $! >pid
 wait $!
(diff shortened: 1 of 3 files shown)


adejanovski:
Hi @rtib, 

I know it's been a while since you posted this PR, but would it be possible to add explanations for this feature in the docs?
Otherwise no one will be aware that this is possible.

Thanks!

rtib:
Sure, I wrote some thought about that and add an explanation with example to minio_setup.md. I've also rebased this to master.

adejanovski (approved):

PR #552, opened 2023-01-04
[k8s] supporting k8s-style secrets stored in a folder

### LINKED ISSUE #493: Allow Medusa to take CQL credentials through a file
The K8ssandra-operator modular secrets backend requires the ability for Medusa to read CQL credentials from files. Currently only environment variables are supported.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1619) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1619
┆priority: Medium

ufoot:
Fixes #493 

 Allows medusa CQL and nodetool passwords to be stored following a "k8s-like" convention, with a `username` and a `password` file stored in a single directory.

Example:

* `/etc/medusa/secrets/cql/username` contains "my-cql-username"
* `/etc/medusa/secrets/cql/password` contains "my-cql-password"
* `/etc/medusa/secrets/nodetool/username` contains "my-nodetool-username"
* `/etc/medusa/secrets/nodetool/password` contains "my-nodetool-password"

Then the caller should export:

```
MEDUSA_CQL_K8S_PASSWORD_PATH=/etc/medusa/secrets/cql
MEDUSA_NODETOOL_K8S_PASSWORD_PATH=/etc/medusa/secrets/nodetool
```

And put the corresponding passwords in the 4 files mentioned above. Hopefully this is the k8s standard way of exporting secrets. In this model it is the responsibility of Medusa to be able to read this k8s "standard way of doing things", and applications, hooks, charts, which populate the filesystem, just follow this model.

Also, this PR comes with small configurability on the Medusa side, in practice k8s passwords are configurable in many ways, but we thought it would be interesting to keep it simple at first, and not overload Medusa with many options. So it is "hardcoded" that username is in `username` and password is in `password`. We could make this even more configurable, but not sure users would immediately benefit for that extra bit of fine-tuning, which comes at a complexity cost.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -10,9 +10,14 @@ Modify it to match your requirements:
 ;config_file = <path to cassandra.yaml. Defaults to /etc/cassandra/cassandra.yaml>
 ;cql_username = <username>
 ;cql_password = <password>
+; When using the following setting there must be files in:
+; - `<cql_k8s_secrets_path>/username` containing username
+; - `<cql_k8s_secrets_path>/password` containing password
+;cql_k8s_secrets_path = <path to kubernetes secrets folder>
 ;nodetool_username =  <my nodetool username>
 ;nodetool_password =  <my nodetool password>
 ;nodetool_password_file_path = <path to nodetool password file>
+;nodetool_k8s_secrets_path = <path to nodetool kubernetes secrets folder>
 ;nodetool_host = <host name or IP to use for nodetool>
 ;nodetool_port = <port number to use for nodetool>
 ;certfile= <Client SSL: path to rootCa certificate>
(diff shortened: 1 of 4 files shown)


adejanovski (changes requested):
Looks good @ufoot !

I just have to ask two small things:

- Could you add the new settings to [this sample config file](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa-example.ini)?
- Could you document the changes in [this doc](https://github.com/thelastpickle/cassandra-medusa/blob/master/docs/Configuration.md) and fill in the newly introduced env variables at the end?

Thanks!

adejanovski:
oh, and don't worry about the failing tests in trunk, they never pass and that's something we need to address.

ufoot:
Added a small notice about "what you should do when using this option" here -> https://github.com/thelastpickle/cassandra-medusa/pull/552/commits/f5f23a15918937d8204fd8e5e6bf1531401d9ed2#diff-287489078ac01282f3a3fb2365324988207f8f07246e5095b524fd1e84fca030R21 Not sure it's required/relevant, after all, some other options which seem more important, generally speaking, do not have this extra bit of description so there's a tradeoff between describing in details and keeping it short. Dunno what's the best here, I can remove those 3 extra lines if needed.

PR #562, opened 2023-02-11
improve ipv6 compatibility (fixes #560)

### LINKED ISSUE #560: Medusa k8ssandra container fails to start in ipv6 only environment
Environment: ipv6 only k8s cluster created with kind (ipFamily: ipv6)
Create k8ssandra cluster with medusa enabled.

Medusa fails with following backtrace:
```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/cassandra/medusa/service/grpc/server.py", line 348, in <module>
    server = Server(config_file_path)
  File "/home/cassandra/medusa/service/grpc/server.py", line 51, in __init__
    self.medusa_config = self.create_config()
  File "/home/cassandra/medusa/service/grpc/server.py", line 80, in create_config
    return load_config(args, config_file)
  File "/home/cassandra/medusa/config.py", line 259, in load_config
    config = parse_config(args, config_file)
  File "/home/cassandra/medusa/config.py", line 245, in parse_config
    config.set('storage', 'fqdn', hostname_resolver.resolve_fqdn())
  File "/home/cassandra/medusa/network/hostname_resolver.py", line 30, in resolve_fqdn
    ip_address_to_resolve = ip_address if ip_address != '' else socket.gethostbyname(socket.getfqdn())
socket.gaierror: [Errno -5] No address associated with hostname
```

Code in hostname_resolver.py tries to use socket.gethostbyname on fqdn which resolves only to ipv6 address.
Because socket.gethostbyname does not support ipv6, this fails.
Code shall be changed to resolve also ipv6.

There is similar issue in cassandra_utils.py where socket.gethostbyname is used on 2 places.
These also shall be fixed(it fails there after first problem is fixed)

tom-code:
Fix to work better in ipv6 only environment (k8ssandra cluster) (fixes #560)

This change tries to call socket.getaddrinfo after socket.gethostbyname fails.
socket.gethostbyname always fails in case there is no ipv4 address(but there can be ipv6 address).

socket.gethostbyname is called first to make sure behavior does not change for dual stack cases.

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -37,6 +37,7 @@ from retrying import retry
 
 from medusa.host_man import HostMan
 from medusa.network.hostname_resolver import HostnameResolver
+from medusa.network.hostname_resolver import resolve_name
 from medusa.nodetool import Nodetool
 from medusa.service.snapshot import SnapshotService
 from medusa.utils import null_if_empty, evaluate_boolean
@@ -157,7 +158,7 @@ class CqlSession(object):
 
     def placement(self):
         logging.debug('Checking placement using dc and rack...')
-        listen_address = socket.gethostbyname(self.cluster.contact_points[0])
+        listen_address = resolve_name(self.cluster.contact_points[0])
         token_map = self.cluster.metadata.token_map
 
         for host in token_map.token_to_host_owner.values():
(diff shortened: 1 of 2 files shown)


adejanovski (approved):
Thanks @tom-code !

PR #567, opened 2023-03-20
Support S3/GCP/Azure storage classes

mohammad-aburadeh:
Medusa does not support specifying the storage class name when uploading backups to S3/GCP/Azure. 
This is very important for many customers as it can help to reduce the storage cost. 

In my case, I manage more than 30 Cassandra clusters. We are storing more than 1000TB of backups in S3. 
If we store those backups in S3 CLACIER_IR storage class, we will reduce the cost to 17%. 

This PR closes: #568


medusa/storage/abstract_storage.py
```diff
@@ -261,3 +261,9 @@ def additional_upload_headers(self):
         child classes.
         """
         return {}
+
+    def get_storage_class(self):
+        if self.config.storage_class is not None:
+            return self.config.storage_class
+        else:
+            return None
```
adejanovski, line 269:
```suggestion
        return self.config.storage_class
```

I guess we can simplify this method since this should be equivalent.

adejanovski (changes requested):
It seems like we're missing a change for the [single part azure uploads](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/azure_blobs_storage/concurrent.py#L129) to configure the storage class there.
I've suggested a simplification as well.

I'll need to run some tests to make sure this is all working as expected as well, but wanted to give you some early feedback already.

Thanks for sending this PR!

mohammad-aburadeh:
Hi @adejanovski 

For aws upload error,  I just fixed it. libcloud assumes the storage class names are lowercase. So, I just converted the class name to lowercase when passing it to libcloud. 

For azure single-part upload. Unfuturnaltly, libcloud does not support specifying storage class name for Azure. 

Regards
Mohammad

mohammad-aburadeh:
Hi @adejanovski 

I just fixed the gcp download issue. 
Could you please review it again and let me know? 

Regards
Mohammad

medusa/storage/google_cloud_storage/gsutil.py
```diff
@@ -88,10 +88,12 @@ def cp(self, *, srcs, dst, max_retries=5, parallel_process_count=1):
...
             ]
 
-        if self._config.storage_class is None:
-            storage_class_options = []
-        else:
-            storage_class_options = ["-s", self._config.storage_class]
+        storage_class_options = []
+
+        # storage class option works with cp upload only
+        if dst.startswith('gs://'):
```
adejanovski, line 94: This is incorrect as you can see [here](https://github.com/thelastpickle/cassandra-medusa/actions/runs/4594351406/jobs/8113516504#step:8:251). dst is a PosixPath object which doesn't have a `startsWith()` method.

Please make sure you've tested the changes that are being pushed otherwise it'll create inertia in the review process. I understand this can be challenging since there are several cloud providers here.
Sadly we cannot enable cloud testing for externally contributed PRs due to how secrets are managed in GitHub Actions (which is no different than other CI systems btw, for security reasons).

adejanovski (changes requested):

medusa/storage/google_cloud_storage/gsutil.py
mohammad-aburadeh: Hi @adejanovski 

Sorry for that. 
I do not have a GCP environment to test my changes. I'll create a test environment soon for my next PRs. 
I've fixed the issue by converting the Posixpath to str and I tested my changes. 

Thanks

adejanovski:
Hi @mohammad-aburadeh,

would you mind rebasing on top of `master` to get the latest fixes for CI?
Thanks!

PR #569, opened 2023-03-24
Storage class test

adejanovski:
Draft PR created for the sake of passing integration tests with all storage backends


rzvoncek:
We now have #777 in, this became obsolete.

PR #571, opened 2023-03-31
Concurrent calls of set_upload_bandwidth cause duplication in ".aws/config"

mohammad-aburadeh:
Hi,

We have faced duplication parameters in the ".aws/config" file because "set_upload_bandwidth" is being called in parallel by multiple threads.

> cassandra@cass-01: ~$ more .aws/config
> [default]
> s3 =
> max_bandwidth = 250MB/s
> [default]
> s3 =
> max_bandwidth = 250MB/s

I have contacted AWS support team, and they said that "aws configure set default.s3.max_bandwidth 250Mb/s" is not thread-safe and It doesn't support multiple calls of this command at the same time.

This command should be executed once in medusa at start-up. No need to execute it by each thread.

I have fixed this in the medusa source code by renaming "set_upload_bandwidth" to "prepare_upload" and executing it only once in the "handle_backup" method. Similar to the "preapre_download" method.

### DIFF
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -177,6 +177,8 @@ def handle_backup(config, backup_name_arg, stagger_time, enable_md5_checks_flag,
         logging.debug("Starting backup preparations with Mode: {}".format(mode))
         storage = Storage(config=config.storage)
         cassandra = Cassandra(config)
+        
+        storage.storage_driver.prepare_upload()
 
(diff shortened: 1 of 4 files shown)


adejanovski (approved):
Looks good to me 👍 
Thanks for making this change, I know that's something that has been bugging a lot of folks.

PR #573, opened 2023-04-10
Correct clean_path execution in restore_node_sstableloader function.

cmvelo:
Correcting a bug where `restore_node_sstableloader` is not cleaning up multiple restore directories.

Note: Looking through the `tests/restore_node.py`, I am not seeing any tests around the `restore_node_sstableloader` function and it looks like the file system tests are mocked out.

However, this fix has been tested in both our PreProduction and Production environments successfully.


adejanovski (approved):
Thanks @cmvelo !

PR #608, opened 2023-08-03
Fix reported time in seconds in metrics when backup are >24h

### LINKED ISSUE #606: Reported time in metrics/logs is wrong when backup > 24h
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=34976925)

We have some backup which can take more than 24h.
In this case, the log is the following:

 ```
[2023-08-02 17:09:33,080] INFO: - Started: 2023-08-01 08:52:06
                        - Started extracting data: 2023-08-01 08:52:19
                        - Finished: 2023-08-02 17:09:33
[2023-08-02 17:09:33,080] INFO: - Real duration: 1 day, 8:17:13.968530 (excludes time waiting for other nodes)
```

However, in metrics.json (with `monitoring_provider = local`), generated from: https://github.com/thelastpickle/cassandra-medusa/blob/95c967fb58e350fc2f20c65a19c11fed11cc6748/medusa/backup_node.py#L307-L337

The `.seconds` conversion only takes the second of the day, not of the full duration.

I think it's better to use `.total_seconds` from `datetime`

```
>>> import datetime
>>> start=datetime.datetime.now() - datetime.timedelta(days = 1, minutes=12)
>>> end=datetime.datetime.now()
>>> duration=end - start
>>> print(duration)
1 day, 0:12:02.583009
>>> print("duration  {}".format(duration))
duration  1 day, 0:12:02.583009
>>> print("duration  {}".format(duration.seconds))
duration  722
>>> print("duration  {}".format(duration.total_seconds()))
duration  87122.583009
```

### PR DESCRIPTION
**Pull request by Skunnyk**
- When backup duration is >24h, the .seconds conversion only convert the second of the day, not the full duration.
- This fix #606

### DIFF
--- a/medusa/backup_cluster.py
+++ b/medusa/backup_cluster.py
@@ -68,9 +68,9 @@ def orchestrate(config, backup_name_arg, seed_target, stagger, enable_md5_checks
 
         logging.debug('Emitting metrics')
 
-        logging.info('Backup duration: {}'.format(backup_duration.seconds))
+        logging.info('Backup duration: {}'.format(backup_duration.total_seconds()))
         tags = ['medusa-cluster-backup', 'cluster-backup-duration', backup_name]
-        monitoring.send(tags, backup_duration.seconds)
+        monitoring.send(tags, backup_duration.total_seconds())
 
         tags = ['medusa-cluster-backup', 'cluster-backup-error', backup_name]
         monitoring.send(tags, 0)
--- a/medusa/restore_cluster.py
+++ b/medusa/restore_cluster.py
@@ -76,9 +76,9 @@ def orchestrate(config, backup_name, seed_target, temp_dir, host_list, keep_auth
 
         logging.debug('Emitting metrics')
 
-        logging.info('Restore duration: {}'.format(restore_duration.seconds))
+        logging.info('Restore duration: {}'.format(restore_duration.total_seconds()))
         tags = ['medusa-cluster-restore', 'restore-duration', backup_name]
-        monitoring.send(tags, restore_duration.seconds)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Skunnyk** [comment]
Woopsie, was tired I guess, used `.total_seconds` instead of calling the function `.total_seconds()` :)

File: medusa/backup_cluster.py [hunk 1]
```diff
@@ -68,9 +68,9 @@ def orchestrate(config, backup_name_arg, seed_target, stagger, enable_md5_checks
 
         logging.debug('Emitting metrics')
 
-        logging.info('Backup duration: {}'.format(backup_duration.seconds))
+        logging.info('Backup duration: {}'.format(backup_duration.total_seconds))
```
adejanovski, line 71: **Question:** shouldn't this be `.total_seconds()` as well?

medusa/restore_cluster.py
```diff
@@ -76,9 +76,9 @@ def orchestrate(config, backup_name, seed_target, temp_dir, host_list, keep_auth
 
         logging.debug('Emitting metrics')
 
-        logging.info('Restore duration: {}'.format(restore_duration.seconds))
+        logging.info('Restore duration: {}'.format(restore_duration.total_seconds))
```
adejanovski, line 79: **Question:** shouldn't this be `.total_seconds()` as well?

medusa/backup_cluster.py
Skunnyk: It should, yes, my bad…

adejanovski (approved):
Looks good now. Thanks @Skunnyk

Issue #501, opened 2022-07-15
Prohibit `/` in both the bucket name and prefix


adejanovski:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13938756)

Backups should fail instantly with a proper error message if the bucket name or the prefix contain a `/`.
Users are tempted to control the location of the backups in the bucket but this isn't compatible with Medusa nor the awscli, but the error messages aren't doing a good job at explaining where the problem is coming from.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1662) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1662
┆priority: Medium

PR #619, opened 2023-08-18
added check for bucket name validation

RachanKaur:
Fixes : #501


burmanm (approved):
The trunk errors don't seem to be from these changes.

PR #622, opened 2023-08-22
[S3 storage] Bring back handling credentials via boto

### LINKED ISSUE #613: Post S3 revamp: bring back multiple auth capabilities using Botocore
The use of boto to deal with the authentication was removed in the S3 refactor work and we need to bring it back so we can easily support AWS roles as well as locally stored credentials.

See [this PR](https://github.com/thelastpickle/cassandra-medusa/pull/442) for reference.

## Definition of Done
- [ ] Botocore is used as dependency to handle getting the AWS credentials
- [ ] AWS roles are supported through Botocore

rzvoncek:
Fixes #613 .


adejanovski (approved):
Looks good :shipit:

PR #623, opened 2023-08-24
[Storage] Adjust chunk sizes based on file_size

### LINKED ISSUE #614: S3: Adjust the chunk size based on the file size
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=36031960)

S3 allows a maximum of 10k chunks per multi part upload. This means the chunk size must be dynamically computed based on the file size to ensure we're staying within that threshold.

The default (and smallest) chunk size is 5 MB and should be recomputed if the file size is larger than 52,428,800 bytes.

The computation should be: 
```
if total_file_size > 50,000,000:
  chunk_size = total_file_size / 9900
else:
  chunk_size = 5,000
```

## Definition of Done
- [ ] chunk size is recomputed dynamically if files larger than 50 GB are uploaded

rzvoncek:
There is a cap on number of chunks per multipart-upload (10k of them). For large files, we need to use bigger chunks to fit within this limit.

Fixes #614.

### DIFF
--- a/medusa/storage/s3_base_storage.py
+++ b/medusa/storage/s3_base_storage.py
@@ -21,9 +21,10 @@ import base64
 import botocore.session
 import collections
 import datetime
-import logging
 import io
 import itertools
+import math
+import logging
 import os
 import subprocess
 import typing as t
@@ -36,8 +37,12 @@ from retrying import retry
 from aiohttp_s3_client import S3Client
 from aiohttp_s3_client.credentials import StaticCredentials
 
-from medusa.storage.abstract_storage import AbstractStorage, AbstractBlob
-
+from medusa.storage.abstract_storage import (
+    AbstractStorage,
+    AbstractBlob,
+    DEFAULT_MULTIPART_PART_SIZE,
+    DEFAULT_MULTIPART_PARTS_COUNT
+)
 
 ManifestObject = collections.namedtuple('ManifestObject', ['path', 'size', 'MD5'])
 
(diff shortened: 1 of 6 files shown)


medusa/storage/s3_base_storage.py
```diff
@@ -261,18 +266,49 @@ async def _upload_small_blob(self, src: str, object_key: str):
...
+        logging.info(f'Calculating part size for file size {file_size} with max parts count {max_parts_cnt}')
+
+        # check if we fit into the parts limit with the default part size
+        num_parts_with_min_part_size = int(math.ceil(file_size / min_part_size))
+        if num_parts_with_min_part_size < max_parts_cnt:
+            part_size, parts_cnt = min_part_size, num_parts_with_min_part_size
+        else:
+            # if not, calculate the part size that will fit into the parts limit
+            part_size = int(math.ceil(file_size / max_parts_cnt))
+            parts_cnt = max_parts_cnt
```
adejanovski, line 304: **nit:** IMO it's fine to use `parts_count` instead of `parts_cnt` for better readability of the code. We don't save much with the contraction.

adejanovski (approved):
Successfully tested, nicely done 👍

rzvoncek:
Thanks for the review. I've rebased on master, fixed a conflict and renamed all `cnt`s to `count`s.

adejanovski:
No need for this anymore following our move to boto3.

PR #630, opened 2023-08-31
Use Boto3 for S3 communications

### LINKED ISSUE #634: Use boto3 in the S3 classes
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=38062246)

After running into issues when uploading large files using the aiohttp-s3-client we decided to move to boto3 which is both officially supported and works perfectly.
The switch is easy to make and should provide a rock solid S3 backend class to Medusa.

## Definition of Done
- [ ] aiohttp-s3-client is removed as a dependency
- [ ] boto3 is used for all interactions with s3(-compatible) backends
- [ ] Large files (> 100GB) are uploaded in a fault resilient fashion with tunable speeds using concurrency settings

adejanovski:
Fixes #634

### DIFF
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -360,7 +360,7 @@ def backup_snapshots(storage, manifest, node_backup, node_backup_cache, snapshot
 
             dst_path = str(node_backup.datapath(keyspace=snapshot_path.keyspace,
                                                 columnfamily=snapshot_path.columnfamily))
-            logging.debug("destination path: {}".format(dst_path))
+            logging.debug("Snapshot destination path: {}".format(dst_path))
 
             manifest_objects = list()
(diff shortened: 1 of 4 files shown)


adejanovski:
This is good to go @rzvoncek, feel free to merge when it suits you.

rzvoncek (approved):

H3199:
Does this support role based authentication?

rzvoncek:
Hi @H3199 . 

If you mean the _EC2 instance role_ based authentication, then yes, that is still supported.

A conventional role-based access control (RBAC) is something we do not support in Medusa. You could probably achieve some of it by generating different credential files to model roles or users, but that's about it.

Issue #628, opened 2023-08-30
Remove reliance on Libcloud/azcli for Azure blob storage backend


adejanovski:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37239969)

As we're only using a small subset of Libcloud's features and it's currently limiting our ability to make Medusa faster and safer, we have decided to move away from it and implement async http operations for multipart uploads.

In that effort, we will be moving the Azure storage classes to rely on [the official Azure storage python library](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/storage/azure-storage-blob) that should handle the following features:

- multipart uploads with automated retries of failed parts
- concurrent uploads

This also allows us to remove:

- The reliance on the azcli (and its dependencies)
- The concurrent code that was built to upload multiple files at the same time

This new implementation will use the same global threshold defined for using multipart uploads. 

Implementing rate limiting of uploads is out of scope for this ticket.

## Definition of Done
- [ ] azcli isn't used anymore, nor any command line tool to upload files to Azure Blob Storage
- [ ] Resilient and concurrent multipart uploads are implemented for the Azure backend

Issue #612, opened 2023-08-16
Post S3 revamp: add KMS headers for upload operations


adejanovski:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=36030027)

After the S3 backend refactor we need to support KMS keys again.
This should be done by adding the extra kms headers to the http PUT operations.
Reads don't need that as the key will be detected automatically.

## Definition of Done
- [ ] KMS headers are sent in the PUT operations
- [ ] Integration tests involving KMS encryption pass

PR #631, opened 2023-08-31
[Storage] Replace gsutil with an aio library

### LINKED ISSUE #627: Remove libcloud/gsutil reliance for GCS
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37239432)

As we're only using a small subset of Libcloud's features and it's currently limiting our ability to make Medusa faster and safer, we have decided to move away from it and implement async http operations for multipart uploads.

In that effort, we will be moving the GCS storage classes to rely on a python library (which still needs to be defined) that should handle the following features:

- multipart uploads with automated retries of failed parts
- concurrent uploads

Current candidates are:
- The official storage python lib from Google: [link](https://github.com/googleapis/python-storage/tree/main)
- gcloud-aio: [link](https://github.com/talkiq/gcloud-aio) (last update date is concerning here)

This also allows us to remove:

- The reliance on the gsutil (and its dependencies)
- The concurrent code that was built to upload multiple files at the same time

This new implementation will use the same global threshold defined for using multipart uploads. 

Implementing rate limiting of uploads is out of scope for this ticket.

## Definition of Done
- [ ] gsutil isn't used anymore, nor any command line tool to upload files to GCS
- [ ] Resilient and concurrent multipart uploads are implemented for the GCS backend

### LINKED ISSUE #637: Retry individual blobs download instead of retrying the whole download list
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=38343509)

Currently the retry annotation is placed on `download_blobs()` which results in redownloading all the files when one fails. We need to move the annotation on the `_download_blob()` method instead, so that we only retry failed files individually.

## Definition of Done
- [ ] for all backends, the retry annotation is placed on `_download_blob()` instead of `download_blobs()`
- [ ] gcs present in #631 
- [ ] azure present in #638 
- [x] s3 already present in [here](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/s3_base_storage.py#L396)

rzvoncek:
There is a bit of duplicated code, but that will go into the `AbstractStorage` once we do #628.
The S3 ITs are failing, because I enforced KMS and there is a KMS test that is not working because #612 is still open.

Oh, and I forgot to remove unused code, will handle asap.

Fixes #627.
Fixes #637.


medusa/backup_node.py
```diff
@@ -31,7 +31,7 @@
 from medusa.index import add_backup_start_to_index, add_backup_finish_to_index, set_latest_backup_in_index
 from medusa.monitoring import Monitoring
 from medusa.storage import Storage, format_bytes_str, ManifestObject, divide_chunks
-from medusa.storage.google_storage import GSUTIL_MAX_FILES_PER_CHUNK
+from medusa.storage.google_storage import GOOGLE_MAX_FILES_PER_CHUNK
```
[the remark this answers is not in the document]
rzvoncek: You're correct, we don't need this. The new storage driver implementations handle this themsleves (eventually we'll do this in the `AbstractStorage` anyway).

medusa/storage/google_storage.py
```diff
@@ -13,106 +13,281 @@
...
+        if src.startswith("gs"):
+            resp = await self.gcs_storage.copy(
+                bucket=self.bucket_name,
+                object_name=f'{src}'.replace(f'gs://{self.bucket_name}/', ''),
+                destination_bucket=self.bucket_name,
+                new_name=object_key,
+            )
+            resp = resp['resource']
+        else:
+            resp = await self.gcs_storage.upload_from_filename(
```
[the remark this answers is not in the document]
rzvoncek: Done.

medusa/storage/google_storage.py
```diff
@@ -13,106 +13,281 @@
...
+        await self.gcs_storage.delete(
+            bucket=self.bucket_name,
+            object_name=obj.name
+        )
 
-    def _download_blobs(self, srcs, dest):
-        with GSUtil(self.config) as gsutil:
-            for parent, src_paths in _group_by_parent(srcs):
-                yield self._download_paths(gsutil, parent, src_paths, dest)
+    @retry(stop_max_attempt_number=MAX_UP_DOWN_LOAD_RETRIES, wait_fixed=5000)
```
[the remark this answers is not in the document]
rzvoncek: Done.

medusa/storage/google_storage.py
```diff
@@ -13,106 +13,286 @@
...
+            '[Storage] Downloading gcs://{}/{} -> {}'.format(
+                self.config.bucket_name, object_key, file_path
+            )
+        )
 
-    def get_object_datetime(self, blob):
-        logging.debug("Blob {} last modification time is {}".format(blob.name, blob.extra["last_modified"]))
-        return parser.parse(blob.extra["last_modified"])
+        try:
+            await self.gcs_storage.download_to_filename(
```
adejanovski, line 272: **Issue:** This method reads the files at once and puts their content into memory. We need to use [download_stream()](https://github.com/talkiq/gcloud-aio/blob/b170cf916f0ad52357bd3457ce5b905dd9170132/storage/gcloud/aio/storage/storage.py#L352C1-L352C1) instead which will return a BufferedStream which should be writeable to a file without loading all the data in memory.

setup.py
```diff
@@ -70,7 +70,8 @@
         'dnspython>=2.2.1',
         'asyncio==3.4.3',
         'aiohttp==3.8.5',
-        'aiohttp-s3-client==0.8.17'
+        'aiohttp-s3-client==0.8.17',
+        'gcloud-aio-storage==8.3.0'
```
adejanovski, line 74: **Note:** I added this dependency which was missing from setup.py

adejanovski (changes requested):
Large files downloads are failing due to the download method being invoked.

medusa/storage/google_storage.py
rzvoncek: Done. Pushed a new commit, together with a rebase on a recent master.

medusa/storage/google_storage.py
```diff
@@ -13,106 +13,292 @@
...
+    def delete_objects(self, objects: t.List[AbstractBlob]):
+        loop = self._get_or_create_event_loop()
+        loop.run_until_complete(self._delete_objects(objects))
 
-        # ensure srcs is always a list
-        if isinstance(srcs, str) or isinstance(srcs, Path):
-            srcs = [srcs]
+    async def _delete_objects(self, objects: t.List[AbstractBlob]):
+        coros = [self._delete_object(obj) for obj in objects]
+        await asyncio.gather(*coros)
```
adejanovski, line 235: **Question:** Is this going to delete all objects concurrently? Maybe our concurrency setting should apply here to avoid sending too many concurrent requests?

medusa/storage/google_storage.py
```diff
@@ -13,106 +13,292 @@
...
-        with GSUtil(self.config) as gsutil:
-            for parent, src_paths in _group_by_parent(srcs):
-                yield self._download_paths(gsutil, parent, src_paths, dest)
+    def download_blobs(self, srcs: t.List[t.Union[Path, str]], dest: t.Union[Path, str]):
+        loop = self._get_or_create_event_loop()
+        loop.run_until_complete(self._download_blobs(srcs, dest))
+
+    async def _download_blobs(self, srcs: t.List[t.Union[Path, str]], dest: t.Union[Path, str]):
+        coros = [self._download_blob(src, dest) for src in map(str, srcs)]
+        await asyncio.gather(*coros)
```
adejanovski, line 250: **Question:** is it downloading all files concurrently at the same time?

medusa/storage/google_storage.py
rzvoncek: Yes, it's going ham doing all at once. The idea was that a download should go as unthrottled as possible because we want it to go through asap. But I don't see any problem chunking this.

medusa/storage/google_storage.py
rzvoncek: Yes, the same as with the deletes. Do we chunk by concucurrent transfers?

adejanovski (approved):
Awesome stuff! :shipit:

PR #638, opened 2023-09-13
[Storage] Replace Azure CLI with python SDK

### LINKED ISSUE #628: Remove reliance on Libcloud/azcli for Azure blob storage backend

### LINKED ISSUE #637: Retry individual blobs download instead of retrying the whole download list

rzvoncek:
Fixes #628.
Fixes #637.


rzvoncek:
Rebased on the recent master.

medusa/storage/azure_storage.py
```diff
@@ -1,81 +1,336 @@
...
             )
+        )
 
-        return driver
+        with open(src, "rb") as data:
+            blob_client = await self.azure_container_client.upload_blob(
+                name=object_key,
+                data=data,
+                overwrite=True,
+                max_concurrency=4,
```
adejanovski, line 183: **Todo:** after doing some benchmarks, it looks like 16 gives us better performance. Could you set this value?

medusa/storage/azure_storage.py
```diff
@@ -1,81 +1,336 @@
...
-            multi_part_upload_threshold=int(self.config.multi_part_upload_threshold)
-        )
-
-    def download_blobs(self, srcs, dest):
-        return medusa.storage.azure_blobs_storage.concurrent.download_blobs(
-            self, srcs, dest, self.bucket,
-            max_workers=self.config.concurrent_transfers,
-            multi_part_upload_threshold=int(self.config.multi_part_upload_threshold)
-        )
+    def _get_or_create_event_loop(self) -> asyncio.AbstractEventLoop:
```
adejanovski, line 323: **Suggestion:** this seems to be repeated in all the storage classes, maybe we could refactor it to have a single version that all storage classes will use? In the abstract storage possibly?

adejanovski (changes requested):
One tiny change request and a refactoring ticket suggestion, then we should be good for a merge.

medusa/storage/azure_storage.py
rzvoncek: Pushed a change.

medusa/storage/azure_storage.py
rzvoncek: It's indeed repeated, I'll be pushing a different PR fixing this.

adejanovski (approved):

PR #640, opened 2023-09-15
Consolidate the storage implementations + Remove libcloud dependency

### LINKED ISSUE #618: Storage: Prevent leaking storage backend connections
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=36147728)

Various storage implementations do not handle connections to the backend correctly. With libcloud, we did care about closing connections.

As we replaced libcloud with the aiohttp-s3-client, it became obvious the connections are leaking. A naive implementation of placing the closing logic in a __del__ method did not work. When running ITs, a fast succession of steps would just call several __init__s of the storage implementation, but somehow forget about the __del__s.

We'll need to spend more time doing this properly. One approach is to use the `with` keyword any time we instantiate a Storage. This way we'll have more control, because calls to __enter__ and __exit__ are more reliable. 

# Definition of done
- [ ] S3 (and compatible) storage backed does not leak connections
- [ ] GCS storage backend does not leak connections
- [ ] Azure storage backend does not leak conenctions

rzvoncek:
Fixes #618 .


rzvoncek:
Also pushed a change to fix the bug reported by SonarCloud.

k8s/Dockerfile-azure
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04 as base
+k8s/Dockerfile-azure FROM ubuntu:18.04 as base
```
rzvoncek, line 1: I'm actally not sure how this sneakedi n.

adejanovski: lol, we're removing this file anyway.

medusa/storage/__init__.py
```diff
@@ -72,39 +66,36 @@ def __init__(self, *, config):
...
+        self.storage_driver.connect()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.storage_driver.disconnect()
+
+    def _load_storage(self):
         logging.debug('Loading storage_provider: {}'.format(self._config.storage_provider))
-        if self._config.storage_provider == Provider.GOOGLE_STORAGE:
+        if self._config.storage_provider.lower() == 'google_storage':
```
adejanovski, line 81: **Suggestion:** we should create a refactoring ticket to create new constants for the storage providers.

medusa/storage/azure_storage.py
```diff
@@ -306,31 +145,76 @@ async def _download_blob(self, src: str, dest: str):
...
 
-    def get_cache_path(self, path: str) -> str:
-        logging.debug(
-            "My cache path is {}".format(path)
+        with open(src, "rb") as data:
+            blob_client = await self.azure_container_client.upload_blob(
+                name=object_key,
+                data=data,
+                overwrite=True,
+                max_concurrency=4,
```
adejanovski, line 187: **Suggestion:** we had settled on 16 ;)

adejanovski (approved):
Fantastic job!
We're good for a merge (with two refactoring tickets that need to be created)

medusa/storage/__init__.py
rzvoncek: https://github.com/thelastpickle/cassandra-medusa/issues/646

medusa/storage/azure_storage.py
rzvoncek: Fixed.

PR #647, opened 2023-09-19
[S3 Storage] Obey throughput limit configuration

### LINKED ISSUE #611: Post S3 backend revamp: handle throttling
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=36029433)

After we've refactored the S3 backend classes to use the new async library, we need to find a safe way to throttle the uploads.

## Definition of Done
- [x] S3 uploads comply with the max throughput setting from Medusa's config
- [x] Throttling is applied at a global level (concurrent uploads summed up throughput should not go over the threshold)

rzvoncek:
Fixes #611.


adejanovski (approved):
Brilliant!

PR #668, opened 2023-10-13
Support backups of DSE with Search

### LINKED ISSUE #659: Add support for DSE search
DSE search uses solr indexes which are stored on disk and need to be backed up additionally to sstables.
We need to add the folders that contain these indexes to the ones that are backed up.

## Definition of Done
- [ ] The Solr indexes are backed up and restored by Medusa

rzvoncek:
Fixes #659 .


rzvoncek:
This change is not doing the right thing.

PR #669, opened 2023-10-18
 Backup & Restore DSE Internals and DSE Search

### LINKED ISSUE #659: Add support for DSE search

### LINKED ISSUE #664: Handle DSE folders correctly
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=40977146)

The DSE comes with a few extra folders in `/var/lib/cassandra` (and probably elsewhere) and Medusa currently cannot handle those.

One specific example is `/var/lib/cassandra/metadata/` which we should treat, probably, in a similar way to `system.local` and `system.peers`.

When working on #659 , I already noticed there's a whole buch of these folders, and I have a `sed` command to modify them. That command is a good starting point to identify these folders and then treat them correctly.

## Definition of done
- [x] Identify all the extra folders of DSE
- [ ] Backup and (not) restore those folders as requested

rzvoncek:
Fixes #659.
Fixes #664.


rzvoncek:
We got an idea to not rebuild the indices on restore, but rather do a "soft" snapshot by pausing the search index while it's being backed up.

PR #674, opened 2023-10-26
Add gRPC config values to configure max send and max receive sizes

### LINKED ISSUE #673: Add the ability to configure gRPC max send / recv sizes
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=42727954)

The GetBackups() gRPC method can return really large result sets if there are many backups available. This will cause the gRPC server to prevent sending that response as we max out the send/recv sizes. Alternatively, allowing modifying these can also be used to limit memory consumption from jumping to unpredicted amounts.

burmanm:
Sets the default values to 512MB for send and 128MB to recv (I don't see why Medusa would receive larger messages in the gRPC layer).

Fixes #673

### DIFF
--- a/medusa/config.py
+++ b/medusa/config.py
@@ -73,7 +73,7 @@ LoggingConfig = collections.namedtuple(
 
 GrpcConfig = collections.namedtuple(
     'GrpcConfig',
-    ['enabled']
+    ['enabled', 'max_send_message_length', 'max_receive_message_length']
 )
 
 KubernetesConfig = collections.namedtuple(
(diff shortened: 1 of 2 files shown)


adejanovski (changes requested):
Looks good but you have some formatting issues which break the build:

```
./medusa/service/grpc/server.py:55:91: E251 unexpected spaces around keyword / parameter equals
./medusa/service/grpc/server.py:55:93: E251 unexpected spaces around keyword / parameter equals
```

burmanm:
Right.. weirdly my flake8 reports 203 errors, so I did miss that one.

adejanovski (approved):

PR #675, opened 2023-10-26
[S3 Storage] Use boto's connection pool so transfers are trully parallel

### LINKED ISSUE #670: Make s3 storage (boto) work with asyncio
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=41924411)

The s3 integration tests are taking too long. Up to 1h compared to ~15-20 min for other storage backends. This is because, despite the medusa storage classes using asyncio, the boto library underneath doesn't. In the end, we end up serialising everything. 

It is really frustrating to keep waiting for the s3 tests on each PR. 

A candidate library to help us here is https://github.com/terrycain/aioboto3.

rzvoncek:
Fixes #670 .


medusa/storage/s3_base_storage.py
```diff
@@ -328,18 +346,29 @@ async def _upload_blob(self, src: str, dest: str) -> ManifestObject:
...
+            'Key': object_key,
+            'Config': self.transfer_config,
+            'ExtraArgs': kms_args,
+        }
+        # we are going to combine asyncio with boto's threading
+        # we do this by submitting the upload into an executor
+        loop = asyncio.get_event_loop()
+        future = loop.run_in_executor(self.executor, self.__upload_file, upload_conf)
+        # and then ask asyncio to yield until it completes
+        mo = await future
```
adejanovski, line 361: **Question:** I fail to understand how that parallelizes uploads. Here you're uploading a single file and then wait for the future (so the upload) to complete, until you move to the next one. Or am I missing something?

rzvoncek: The way I suspect this works is that the `await` keyword _yields_ the execution of the current coroutine. 

The `_upload_object` is async already, which means several of them happen concurrently. The issue was they didn't because the call to s3_client.upload() was blocking.

With the change here, we hide the blocking call to client into a future, and the yield until it completes.

adejanovski (approved):

PR #681, opened 2023-11-15
Make gRPC Service's BackupStatus return a node backup

### LINKED ISSUE #660: Make gRPC BackupStatus() operation be about the current node only.
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=40584156)

[The gGRPC's BackupStatus()](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/service/grpc/server.py#L166) currently works atop cluster-wide backups. Considering we use the this in the context of the k8ssandra-operator, where we need to determine a status of a single/current node, this is inaccurate.

We need to make the `BackupStatus()` return a status of a specific backup for the current node.

So instead of:
```
backup = storage.get_cluster_backup(request.backupName)
```

We need to do something like:
```
backup = storage.get_node_backup(storage_config.fqdn, request.backupName)
```

We should also call `update_backup_status` from `set_backup_result` so that the status is updated in `BackupMan` when the backup finishes or fails.

rzvoncek:
Fixes #660 .

### DIFF
--- a/medusa/service/grpc/medusa.proto
+++ b/medusa/service/grpc/medusa.proto
@@ -44,12 +44,9 @@ message BackupStatusRequest {
 }
 
 message BackupStatusResponse {
-  repeated string  finishedNodes = 1;
-  repeated string  unfinishedNodes = 2;
-  repeated string  missingNodes = 3;
-  string           startTime = 4;
-  string           finishTime = 5;
-  StatusType status = 6;
+  string           startTime = 1;
+  string           finishTime = 2;
+  StatusType       status = 3;
 }
 
 message DeleteBackupRequest {
--- a/medusa/service/grpc/medusa_pb2.py
+++ b/medusa/service/grpc/medusa_pb2.py
@@ -13,7 +13,7 @@ _sym_db = _symbol_database.Default()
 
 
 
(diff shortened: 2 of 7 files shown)


medusa/service/grpc/medusa.proto
```diff
@@ -42,9 +42,12 @@ message BackupStatusRequest {
 }
 
 message BackupStatusResponse {
+  // this response is about a _node_ backup. The fields about nodes are not correct here
+  // we should deprecate/remove them but that breaks backwards compatibility
```
emerkle826, line 46: Thought:
Could we introduce a new response and deprecate the old one? If that's feasible, maybe create a refactor ticket.

emerkle826 (approved):

medusa/service/grpc/medusa.proto
rzvoncek: Yes, I agree with this. We initially wanted to preserve backwards compatibility.
But I checked with @adejanovski and we decided to just drop this straight away.
I've pushed an ammended commit that drops this (and it also does some cleanup after rebasing on top of the other PRs).

rzvoncek:
I'll go ahead and close this. The k8ssandra e2e test is broken, but I'm going to fix it in #686 .

PR #714, opened 2024-02-19
Add mtls to mgmt api connection

burmanm:
* Add new config parameters: ca_cert, tls_cert and tls_key. 
* Use request…s persistant sessions on Medusa mgmt-api communication
* Set TLS properties if tls_cert config is set (support both TLS and mTLS setup)


.github/workflows/ci.yml
```diff
@@ -294,6 +294,10 @@ jobs:
         mv tests/integration/.coverage .
         coverage xml
 
+    - name: Setup tmate session
```
adejanovski, line 297: **Suggestion:** this could be removed now

adejanovski (approved):

PR #720, opened 2024-03-07
Make grpc's AsynBackup trully Async

rzvoncek:
Fixes https://github.com/k8ssandra/k8ssandra-operator/issues/1183
Fixes https://github.com/k8ssandra/k8ssandra-operator/issues/1231
Fixes https://github.com/thelastpickle/cassandra-medusa/issues/721

The GRPC's AsyncBackup endpoint was not async. It was blocking until the backup finished. This PR fixes that, and then adjusts the ITs so that they work with the new async implementation. 
I've also touched the logging during ITs so that we don't have thousands of lines about Cassandra not being yet available.
I've also touched how we build the Medusa container image for k8s; it now has the Medusa entry point on the system path again.

Aside from the ITs, I've tested this by:
- manually hitting the (Async)Backup endpoints with Bloom RPC
- loading a Medusa image built off this branch into a k8s cluster with k8ssandra-operator and creating a backup
- observing that the backups happen in parallel
- observing that the MedusaBackupJob gets its status updated after a node finishes (or fails) instead of waiting till all the nodes are done.

### DIFF
--- a/k8s/Dockerfile
+++ b/k8s/Dockerfile
@@ -31,7 +31,7 @@ ENV POETRY_VIRTUALENVS_IN_PROJECT=true
 RUN python3 -m pip install -U pip && pip3 install --ignore-installed --user poetry
 
 # Build medusa itself so we can add the executables in the final image
-RUN cd /build && poetry install
+RUN cd /build && poetry build && poetry install
 
 # Could be python:slim, but we have a .sh entrypoint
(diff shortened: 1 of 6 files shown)


adejanovski (approved):
Brilliant! I had tried before and couldn't figure out how to make the grpc server async.
Nicely done!

PR #732, opened 2024-03-28
Add option to prefer incomplete backups when doing a purge

### LINKED ISSUE #607: max_backup_count should keep valid backups
If for any reasons your backups fails, the `medusa purge` task will delete what you have configured in `max_backup_count`, without taking the status in account.
So you can end up with only incompletes backups:

`max_backup_count = 5`: 
```
$ medusa list-backups
[2023-07-26 17:41:44,995] INFO: Found credentials in shared credentials file: /etc/medusa/credentials-s3
scylladb01-production-20230719-1100 (started: 2023-07-19 11:00:40, finished: Incomplete [0 of 6 nodes finished])
scylladb01-production-20230721-1100 (started: 2023-07-21 11:01:24, finished: Incomplete [0 of 6 nodes finished])
scylladb01-production-20230723-1100 (started: 2023-07-23 11:00:55, finished: Incomplete [0 of 6 nodes finished])
scylladb01-production-20230724-1100 (started: 2023-07-24 11:01:20, finished: Incomplete [0 of 6 nodes finished])
scylladb01-production-20230725-1100 (started: 2023-07-25 11:00:43, finished: Incomplete [0 of 6 nodes finished])
```

In this case, we don't have any valid backup due to the purge, only incomplete ones.

We may be able to keep `max_backup_count` valid backups after a purge. Not sure how to implement that btw.

Yes I know, the real solution is to avoid failed backups :D

rzvoncek:
Fixes #607.

### DIFF
--- a/medusa/medusacli.py
+++ b/medusa/medusacli.py
@@ -332,12 +332,15 @@ def build_index(medusa_config, noop):
 
 
 @cli.command(name='purge')
+@click.option('--prefer-incomplete', default=False, is_flag=True,
+              help='Prioritise incomplete backups when purging by count')
 @pass_MedusaConfig
-def purge(medusaconfig):
+def purge(medusaconfig, prefer_incomplete):
     """
     Delete obsolete backups
     """
     medusa.purge.main(medusaconfig,
+                      prefer_incomplete=prefer_incomplete,
                       max_backup_age=int(medusaconfig.storage.max_backup_age),
                       max_backup_count=int(medusaconfig.storage.max_backup_count))
 
--- a/medusa/purge.py
+++ b/medusa/purge.py
@@ -27,21 +27,29 @@ from medusa.monitoring import Monitoring
 from medusa.storage import Storage, format_bytes_str
 
 
-def main(config, max_backup_age=0, max_backup_count=0):
+def main(config, prefer_incomplete=False, max_backup_age=0, max_backup_count=0):
     backups_to_purge = set()
     monitoring = Monitoring(config=config.monitoring)
 
     try:
         logging.info('Starting purge')
         with Storage(config=config.storage) as storage:
-            # Get all backups for the local node
-            logging.info('Listing backups for {}'.format(config.storage.fqdn))
+
+            # Pre-fetch the backup index
             backup_index = storage.list_backup_index_blobs()
-            backups = list(storage.list_node_backups(fqdn=config.storage.fqdn, backup_index_blobs=backup_index))
-            # list all backups to purge based on date conditions
-            backups_to_purge |= set(backups_to_purge_by_age(backups, max_backup_age))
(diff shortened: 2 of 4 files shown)


medusa/medusacli.py
```diff
@@ -332,12 +332,15 @@ def build_index(medusa_config, noop):
 
 
 @cli.command(name='purge')
+@click.option('--prefer-incomplete', default=False, is_flag=True,
+              help='Prioritise incomplete backups when purging by count')
```
adejanovski, line 336: **Question:** in which case do you think we may prefer to keep incomplete backups? They cannot be restored anyway so it doesn't seem like something we'd want to keep.

medusa/purge.py
```diff
@@ -62,6 +70,58 @@ def main(config, max_backup_age=0, max_backup_count=0):
         sys.exit(1)
 
 
+def find_backups_to_purge(
+        complete_cluster_backup_names, node_backups, prefer_incomplete, max_backup_age, max_backup_count
+):
+    backups_to_purge = set()
+
+    # we purge old backups always
+    backups_to_purge |= set(backups_to_purge_by_age(node_backups, max_backup_age))
```
adejanovski, line 79: **Question:** So we're considering that incomplete backups shouldn't be affected by the new purge priorities that we're setting? How is that purge rule different than the purge by count rule?

medusa/purge.py
```diff
@@ -62,6 +70,58 @@ def main(config, max_backup_age=0, max_backup_count=0):
...
+
+    # if we already keep only max_backup_count or less, we return
+    if max_backup_count == 0 or len(node_backups) - len(backups_to_purge) <= max_backup_count:
+        return backups_to_purge
+
+    # if we don't care about the incomplete ones, we purge by count and return
+    if not prefer_incomplete:
+        backups_to_purge |= set(backups_to_purge_by_count(node_backups, max_backup_count))
```
adejanovski, line 87: **Issue:** Unless I'm missing something, I think this is doing the opposite of what was requested in the ticket.
If we don't prefer incomplete backups (which should mean we prefer complete backups?), then you're applying the same rules as before here, which can lead to having only incomplete backups left after the purge.
The implementation we're looking for here is to purge incomplete backups first (disregarding their age and count) and only then apply the age/count purge rules.

I think we could widely simplify this code and the rules by always purging incomplete backups in priority, and only then apply the purge rules.
wdyt?

adejanovski (changes requested):
I'm a little confused with this implementation and how it drifts from what was originally asked in the ticket.
Let's connect to chat about this.

adejanovski:
Bump @rzvoncek 👋

rzvoncek:
Closing in favour of https://github.com/thelastpickle/cassandra-medusa/pull/778

Issue #543, opened 2022-11-24
Allow for optional SSL CERT verification on SSL enabled Cassandra clusters


alier350:
Currently SSL CERT verification is enabled in code https://github.com/thelastpickle/cassandra-medusa/blob/3991da1a5272f8e26b668ac6cbff737cf705677e/medusa/cassandra_utils.py#L75 However we noticed that this check fails on wildcard certificates with 2 nested sub-domains (which is expected according to [RFC2818](https://www.rfc-editor.org/rfc/rfc2818#section-3.1) spec), though an option to control this functionality as a CLI parameter , i.e. `--ssl-verification yes` or configuration in medusa.ini would really help to have SSL enabled Cassandra clusters without strict certificate verification for large Cassandra estates where a wildcard certificate is easier to produce and distribute rather than signing every node individually.

adejanovski:
Hi @alier350,

this sounds like a fair request.
You can send a PR if you wish to have this available sooner rather than later.

Thanks!

rzvoncek:
This might relate to https://github.com/thelastpickle/cassandra-medusa/issues/563

PR #736, opened 2024-04-05
Enable SSL without client verification

### LINKED ISSUE #563: Support SSL without cert files
Hi, 

We have production Cassandra clusters where SSL (client_encryption_options) is enabled with no validation is required (require_client_auth: false). 

But medusa does not support enabling SSL without cert file. 

The current connection code in medusa is: 

`

      if self._cassandra_config.certfile is not None:
            ssl_context = SSLContext(PROTOCOL_TLSv1_2)
            ssl_context.load_verify_locations(self._cassandra_config.certfile)
            ssl_context.verify_mode = CERT_REQUIRED
            if self._cassandra_config.usercert is not None and self._cassandra_config.userkey is not None:
                ssl_context.load_cert_chain(
                    certfile=self._cassandra_config.usercert,
                    keyfile=self._cassandra_config.userkey)
            self._ssl_context = ssl_context
`

As you see in the above code, SSL is enabled if the "certfile" is provided. 
But in our case SSL validation is not enabled. 
We need a way to connect to the cluster without a certfile. 

Simply the code should be like this: 

`

      if self._cassandra_config.ssl == True: 
          ssl_context = SSLContext(PROTOCOL_TLSv1_2)

      if self._cassandra_config.certfile is not None:
            
            ssl_context.load_verify_locations(self._cassandra_config.certfile)
            ssl_context.verify_mode = CERT_REQUIRED
            if self._cassandra_config.usercert is not None and self._cassandra_config.userkey is not None:
                ssl_context.load_cert_chain(
                    certfile=self._cassandra_config.usercert,
                    keyfile=self._cassandra_config.userkey)
            self._ssl_context = ssl_context
`

Could you please help to support this in medusa ?? 

Regards
Mohammad

mohammad-aburadeh:
### Description
This PR is to support SSL connection without cert files required. This is needed when "require_client_auth" is disabled in Cassandra configurations. 

### Status

- done, ready for review

Closes #563 and #543

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -74,8 +74,8 @@ class CqlSessionProvider(object):
                                                   password=self._cassandra_config.cql_password)
             self._auth_provider = auth_provider
 
-        if ((self._cassandra_config.ssl is not None and evaluate_boolean(self._cassandra_config.ssl)) or 
-            self._cassandra_config.certfile is not None):
+        if ((self._cassandra_config.ssl is not None and evaluate_boolean(self._cassandra_config.ssl))
+                or self._cassandra_config.certfile is not None):
             ssl_context = SSLContext(PROTOCOL_TLSv1_2)
 
             if self._cassandra_config.certfile is not None:
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -74,8 +74,8 @@ class CqlSessionProvider(object):
                                                   password=self._cassandra_config.cql_password)
             self._auth_provider = auth_provider
 
-        if ( (self._cassandra_config.ssl is not None and evaluate_boolean(self._cassandra_config.ssl)) or 
-            self._cassandra_config.certfile is not None ):
+        if ((self._cassandra_config.ssl is not None and evaluate_boolean(self._cassandra_config.ssl)) or 
+            self._cassandra_config.certfile is not None):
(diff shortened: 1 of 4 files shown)


mohammad-aburadeh:
Hi @rzvoncek 
This is requested by many customers. Can you plz review it ?

medusa/cassandra_utils.py
```diff
@@ -74,15 +74,18 @@ def __init__(self, ip_addresses, config):
                                                   password=self._cassandra_config.cql_password)
             self._auth_provider = auth_provider
 
-        if self._cassandra_config.certfile is not None:
+        if ((self._cassandra_config.ssl is not None and evaluate_boolean(self._cassandra_config.ssl))
+                or self._cassandra_config.certfile is not None):
             ssl_context = SSLContext(PROTOCOL_TLSv1_2)
```
rzvoncek, line 79: Hi @mohammad-aburadeh! Sorry for the delay, but I finally managed to find the time to look at this.

And I'm sorry to admit, but I don't quite understand what the proposed change is. If I'm reading this right, the new behaviour is that if the new `ssl` option is present and set to True, it'll just set the `ssl_context` variable ... and then do nothing else. 
The line 88 is indented under the condition of certfile being present, which is basically the old behaviour.

Please, could you clarify what your intent was?

rzvoncek (changes requested):
Hello @mohammad-aburadeh , I'm sorry to send this back to you, but I don't think the change is complete.

PR #737, opened 2024-04-07
Support (S3, GCP, Azure) storage classes

### LINKED ISSUE #568: Allow specifying storage class name when uploading backups to S3
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=23371658)

Medusa does not support specifying the storage class name when uploading backups to S3/GCP/Azure.
This is very important for many customers as it can help to reduce the storage cost.

In my case, I manage more than 30 Cassandra clusters. We are storing more than 1000TB of backups in S3.

mohammad-aburadeh:
Medusa does not support specifying the storage class name when uploading backups to S3/GCP/Azure.
This is very important for many customers as it can help to reduce the storage cost.

Closes #568

### DIFF
--- a/medusa-example.ini
+++ b/medusa-example.ini
@@ -63,7 +63,7 @@ storage_provider = <Storage system used for backups>
 storage_class = <Storage Class Name used to store backups>
 ; AWS S3 Storage classes: STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | INTELLIGENT_TIERING | GLACIER | DEEP_ARCHIVE | GLACIER_IR
 ; GCP Storage classes: STANDARD | NEARLINE | COLDLINE | ARCHIVE
-; AZURE Storage classes: Cool | Hot | Archive
+; AZURE Storage classes: ARCHIVE | COLD | COOL | HOT
 ; https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html
 ; https://cloud.google.com/storage/docs/storage-classes
 ; https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview
--- a/medusa-example.ini
+++ b/medusa-example.ini
@@ -60,6 +60,14 @@ storage_provider = <Storage system used for backups>
 ; storage_provider should be either of "local", "google_storage", "azure_blobs" or the s3_* values from
 ; https://github.com/apache/libcloud/blob/trunk/libcloud/storage/types.py
 
+storage_class = <Storage Class Name used to store backups>
+; AWS S3 Storage classes: STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | INTELLIGENT_TIERING | GLACIER | DEEP_ARCHIVE | GLACIER_IR
+; GCP Storage classes: STANDARD | NEARLINE | COLDLINE | ARCHIVE
+; AZURE Storage classes: Cool | Hot | Archive
+; https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html
+; https://cloud.google.com/storage/docs/storage-classes
+; https://learn.microsoft.com/en-us/azure/storage/blobs/access-tiers-overview
+
 ; Name of the bucket used for storing backups
 bucket_name = cassandra_backups
 
--- a/medusa/storage/azure_storage.py
+++ b/medusa/storage/azure_storage.py
@@ -115,6 +115,7 @@ class AzureStorage(AbstractStorage):
             name=object_key,
             data=data,
             overwrite=True,
+            standard_blob_tier=self.get_storage_class(),
         )
         blob_properties = await blob_client.get_blob_properties()
         return AbstractBlob(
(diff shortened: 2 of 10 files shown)


medusa/storage/google_storage.py
```diff
@@ -122,12 +122,16 @@ async def _upload_object(self, data: io.BytesIO, object_key: str, headers: t.Dic
...
+
+        storage_class = self.storage.get_storage_class(),
+        ex_header = {"storageClass": storage_class}
         resp = await self.gcs_storage.upload(
             bucket=self.bucket_name,
             object_name=object_key,
             file_data=data,
             force_resumable_upload=True,
             timeout=-1,
+            headers=ex_header,
```
rzvoncek, line 134: I could not get this to work with GCS. All the uploads I did ended up with the Standard storage class. Swithich the default bucket storage class mode to `Managed with Autoclass` did not help either.

It's like the GCS ignores the HTTP header. Even if I set it in the request, the response comes with ` 'storageClass': 'STANDARD',`

medusa-example.ini
```diff
@@ -60,6 +60,14 @@ storage_provider = <Storage system used for backups>
 ; storage_provider should be either of "local", "google_storage", "azure_blobs" or the s3_* values from
 ; https://github.com/apache/libcloud/blob/trunk/libcloud/storage/types.py
 
+storage_class = <Storage Class Name used to store backups>
+; AWS S3 Storage classes: STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | INTELLIGENT_TIERING | GLACIER | DEEP_ARCHIVE | GLACIER_IR
```
rzvoncek, line 64: Starting from `GLACIER`, things start to break.
The issue is that we can upload stuff with this storage class, but the we try to read stuff back, and we cannot . At least not in the current implementation.
I think the easy way out is to simply not support the 3 lowest tiers. We 'd need to document this somewhere, and then add checks in the config.py to not allow specifying these.

medusa/storage/azure_storage.py
```diff
@@ -115,6 +115,7 @@ async def _upload_object(self, data: io.BytesIO, object_key: str, headers: t.Dic
             name=object_key,
             data=data,
             overwrite=True,
+            standard_blob_tier=self.get_storage_class(),
```
rzvoncek, line 118: I'm sorry but this does not work. The Azure library needs to have the instances of `StandardBlobTier` clsss (which are just an enum) here.
So we need to do two things:
- on line 28 import the class (`from azure.storage.blob import BlobProperties, StandardBlobTier`)
- map the config values to the values of the enum and pass the whole enum (down the call stack, the library does call `.value()` on the enum.

rzvoncek (changes requested):
Hello @mohammad-aburadeh and sorry to keep you waiting for so long. I've finally managed to have a proper look at this and tried to test it out. Sadly, I have bad news.

In the context of S3, we need to limit the sotrage classes we support. It doesn't seem to be possible to read from Glacier and below, so we should not let Medusa write those because it won't be able to read after itself.

In the context of GCS, the explicit storage classes don't seem to work at all. The header seems just ignored. I didn't find a way to do this aside from setting a default value on the entire bucket.

In the context of Azure, we need to pass enums, not strings, because that's what the client we use expects.

Because of this, I'm sorry to return the PR to you and kindly ask you to do one more iteration to fix/improve things.

federicobaldo:
hi everyone, any help needed here? this feature is very interesting and would like to try using it ASAP

kantipudipythian:
Hi,
We tried the following steps to add the storage_class. We are using AWS S3 bucket for storing the backup files.

1. Upgraded the Medusa version to 0.21.0 from 0.17.2.
2. Added the storage_class parameter to the medusa.ini file.
3. We updated config.py, abstract_storage.py, s3_base_storage.py files accordingly.
4. Ran differential backup.

The backup was successful. But it is taking 1 hour to complete. The previous backups would finish within 2-5minutes. We observed that the manifest.json file is taking more time. Can you please let us know what might be the issue?

rzvoncek:
I've implemented the suggested changes and added integration tests over at https://github.com/thelastpickle/cassandra-medusa/pull/777/checks

kantipudipythian:
Hi,

Below are the tests done on the cluster:

Old Medusa Version: 0.17.2
New Medusa Version: 0.21.0
Storage_class: STANDARD_IA

Test1 (New Medusa version):
New bucket,
storage_class parameter in the medusa.ini file
Started backup.
The 2nd backup took the same time as the first backup around 50 minutes.

Test2 (old Medusa version):
New bucket,
storage_class parameter in the medusa.ini file
Started backup.
The backup was successful, 1st backup took 50 minutes to complete. The 2nd backup was done within 1minute.

Test3 (old Medusa version):
New bucket,
storage_class parameter in the medusa.ini file
Modified config.py, abstract_storage.py, s3_base_storage.py 
Started backup.
The backup was successful, 1st backup took 6 minutes to complete. The 2nd backup was done within 1minute. (There are few backups in the bucket while taking this backup.)

Test4 (old Medusa version):
New bucket,
storage_class parameter in the medusa.ini file
Modified config.py, abstract_storage.py, s3_base_storage.py 
Started backup with the empty new bucket.
The backup was successful, 1st backup took 50mins to complete. The 2nd backup was done within 1minute.

Can you please let us know why the New Medusa version with STANDARD_IA is taking more time?

Thanks,
Kanthi Rekha.

PR #745, opened 2024-04-11
Prevent gRRPC's BackupStatus from canceling pending futures

### LINKED ISSUE #742: gRPC server's BackupStatus cancels AsyncBackup's future
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=59410719)

We've recently made the [AsyncBackup really Async](https://github.com/thelastpickle/cassandra-medusa/commit/7c83424dac0baff75846669d31aa05a04f22dee8), and then went and ["fixed" the BackupStatus endpoint](https://github.com/thelastpickle/cassandra-medusa/commit/49f88ea06406b42d9e8acdd18ef75e3bbf1e200c).

However, this fix introduced a bug. Calling the [register_backup](https://github.com/thelastpickle/cassandra-medusa/commit/49f88ea06406b42d9e8acdd18ef75e3bbf1e200c#diff-0af9023e5dba3232ce866a7482e0b302d746bfa87c8040bb6c3516d481767679R192) in the status endpoint is not correct. Re-registering the backup means the previously registered backup gets removed, and it's pending future cancelled. 

This means that the running instance of grpc server (and the backupman specifically) will never get to handle completion of the future.

The backup itself is not cancelled. I've observed this on a cluster, where an exception was thrown:
```
[2024-04-11 09:00:16,131] INFO: Recording async backup information.
[2024-04-11 09:00:16,131] ERROR: Exception in callback record_backup_info(<Future cancelled>) at /home/cassandra/medusa/service/grpc/server.py:346
handle: <Handle record_backup_info(<Future cancelled>) at /home/cassandra/medusa/service/grpc/server.py:346>
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/home/cassandra/medusa/service/grpc/server.py", line 349, in record_backup_info
    if future.exception():
asyncio.exceptions.CancelledError: Removal of backup requested. Cancelling backup Name: backup-2024-04-11t08-59-44-564z with done state: False
```
but the backup continued and succeeded nonetheless.

We need to fix this, probably by not registering, but "re-registering", which will should probably be a new method that just registers with backupman, but does not cancel any futures.

rzvoncek:
Fixes #742.

Prior to this patch, I could even reproduce the exception from the issue in the integration tests:
```
[2024-04-11 13:17:02,362] INFO: Registered backup id grpc_backup_23
[2024-04-11 13:17:02,363] INFO: Recording async backup information.
[2024-04-11 13:17:02,363] ERROR: Exception in callback record_backup_info(<Future cancelled>) at /Users/zvo/github/cassandra-medusa/medusa/service/grpc/server.py:346
handle: <Handle record_backup_info(<Future cancelled>) at /Users/zvo/github/cassandra-medusa/medusa/service/grpc/server.py:346>
  File "/Users/zvo/anaconda3/lib/python3.11/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/zvo/github/cassandra-medusa/medusa/service/grpc/server.py", line 349, in record_backup_info
    if future.exception():
       ^^^^^^^^^^^^^^^^^^
asyncio.exceptions.CancelledError: Removal of backup requested. Cancelling backup Name: grpc_backup_23 with done state: False
```


medusa/backup_manager.py
```diff
@@ -99,7 +99,7 @@ def set_backup_future(backup_name, future):
 
     # Sets the future for a backup; unknown on overall status at this point unless already existing
     @staticmethod
-    def register_backup(backup_name, is_async):
+    def register_backup(backup_name, is_async, overwrite_existing=True):
```
burmanm, line 102: question: In which case is the overwrite_existing=True and is_async=True ? I can't find such invocation in current codebase so I was wondering is the overwrite_existing=False intended to be used only with async cases?

smutel:
Hello,

I have the same issue. Do you have any workaround to suggest me ? The backups seems completely hanged ...

Thanks.

medusa/backup_manager.py
```diff
@@ -110,8 +110,9 @@ def register_backup(backup_name, is_async):
 
             if backup_name in BackupMan.__instance.__backups:
                 logging.debug("Registered backup name {} found existing, replacing with new".format(backup_name))
```
adejanovski, line 112: **Issue:** This debug logging is inaccurate. We're only replacing with new if `overwrite_existing` is `True`, which is only checked on the next line.

medusa/service/grpc/server.py
```diff
@@ -189,7 +189,7 @@ def BackupStatus(self, request, context):
                     response.finishTime = datetime.fromtimestamp(backup.finished).strftime(TIMESTAMP_FORMAT)
                 else:
                     response.finishTime = ""
-                BackupMan.register_backup(request.backupName, is_async=False)
+                BackupMan.register_backup(request.backupName, is_async=False, overwrite_existing=False)
```
adejanovski, line 192: **Question:** Why are we registering the backup in `BackupStatus()`? Is there a case where we could check a backup that wasn't registered yet?

adejanovski (changes requested):

adejanovski (changes requested):
One logging is inaccurate now and I would like to understand why we're re-registering backups when checking their status.

medusa/backup_manager.py
rzvoncek: Agreed. I've simply removed the line.

medusa/service/grpc/server.py
rzvoncek: We need to do a register here because later, at [line 200](https://github.com/thelastpickle/cassandra-medusa/pull/745/files#diff-0af9023e5dba3232ce866a7482e0b302d746bfa87c8040bb6c3516d481767679L200), we call the `record_status_in_response`, which then calls `BackupMan.get_backup_status(backup_name)`. 

If we are a fresh gRPC server (for example a pod that got restarted) and the first thing we do is to answer a backup status request, this would break. It would break because the BackupMan is an in-memory cache of sorts for backup statuses and if we're not the same instance/process that was up when the backup happened, we don't have the in-memory information.

I realise how clunky this is. I barely understand what the BackupMan does, I haven't yet gotten to understand the why. I would not oppose removing it from the code altogether, unless there's a critical function that it does and that I don't yet know about.

adejanovski (approved):

PR #763, opened 2024-05-16
Make read timeout configurable

### LINKED ISSUE #762: Make S3 read timeout configurable
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=63118978)

This is a feature [requested](https://discord.com/channels/836217371453685760/1237249503296360488/1239907775816011777) by a user at discord.

rzvoncek:
Fixes #762

### DIFF
--- a/medusa/storage/google_storage.py
+++ b/medusa/storage/google_storage.py
@@ -49,6 +49,8 @@ class GoogleStorage(AbstractStorage):
 
         logging.getLogger('gcloud.aio.storage.storage').setLevel(logging.WARNING)
 
+        self.read_timeout = int(config.read_timeout)
+
         super().__init__(config)
 
     def connect(self):
@@ -94,7 +96,8 @@ class GoogleStorage(AbstractStorage):
             # fetch a page
             page = await self.gcs_storage.list_objects(
                 bucket=self.bucket_name,
-                params=params
+                params=params,
+                timeout=self.read_timeout,
             )
 
             # got nothing, return from the function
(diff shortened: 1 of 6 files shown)


medusa/storage/s3_base_storage.py
```diff
@@ -136,7 +136,8 @@ def connect(self):
             region_name=self.credentials.region,
             signature_version='v4',
             tcp_keepalive=True,
-            max_pool_connections=max_pool_size
+            max_pool_connections=max_pool_size,
+            read_timeout=int(self.config.read_timeout),
```
adejanovski, line 140: **Question:** is there a similar feature for GCP and Azure? It would be nice to apply the same settings for all 3.

adejanovski: @rzvoncek ?

rzvoncek: Yes, they are available. I've added commits applying the read timeout config value to Azure and GCP read operations.

adejanovski (approved):

medusa/storage/google_storage.py
```diff
@@ -151,7 +154,7 @@ async def _download_blob(self, src: str, dest: str):
             stream = await self.gcs_storage.download_stream(
                 bucket=self.bucket_name,
                 object_name=object_key,
-                timeout=-1,
+                timeout=self.read_timeout if self.read_timeout is not None else -1,
```
mlallaouret, line 157: Hello @rzvoncek,

I am a user of the k8ssandra operator on gcp and this change can make a breaking change in the restore case if the files are big. We are switching from infinite to 60 seconds timeout if I am not mistaken.

Do you think we should rollback the change for the gcp case or should I post a comment here https://github.com/k8ssandra/k8ssandra-operator/issues/1353 to know if we know when this will be work on ?

PR #791, opened 2024-07-11
Support partially backup

### LINKED ISSUE #116: Support for partial backups: backup only specific keyspaces/tables, etc.
It would be really useful to support partial backups of the data, like:

- backup only specific keyspaces or tables (white list);
- exclude specific keyspaces or tables (black list).

This would greatly help in some cases, when cluster contains some data that don't have much value (like temporary data), could be replayed from other sources, or backup could be used for offloading of "cold/inactive" data that could be restored later on table-by-table decision.

P.S. it's similar to #64, but this functionality is really requested by customers, and supported in other products, like, OpsCenter (keyspace-level only).

dispensable:
support exclude/include fqtn by regex on cli.

for example: backup fqtn like: `test_ks.test_table-dd6ec5d02f9d11ef8b8f13b0bf265570` can be exclude or include with regex `test_ks\.test_table.+`

fix https://github.com/thelastpickle/cassandra-medusa/issues/116

### DIFF
--- a/docs/Performing-backups.md
+++ b/docs/Performing-backups.md
@@ -30,6 +30,8 @@ Options:
                                   Number of concurrent synchronous (blocking)
                                   ssh sessions started by pssh
 
+  --exclude TEXT                  Regex to exclude <keyspace>.<table>
+  --include TEXT                  Regex to include <keyspace>.<table>
   --help                          Show this message and exit.
 ```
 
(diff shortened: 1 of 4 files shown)


Kiiv:
Hi,

I'm wondering why this PR was not merged back in 2024. Is there a technical limitation to it or it was just a project decision not to include this feature in the product ?

Do you think it worth working on conflict resolution  to merge it ?

PR #796, opened 2024-08-14
Make remote command shell act as a login shell

pstef:
This commit adds a parameter to the run_command() method call. The value differs from the pssh's default '$SHELL -c' by adding '-l' to make the shell act as a login shell.

This results in loading .bash_profile or .profile, depending on which shell it is and other circumstances.

The reason to load a profile file is to set up environment, for example PATH.

This is especially handy in environments where Python virtual envs are used. There, to use one, the user has to activate the env by sourcing a file.

### DIFF
--- a/medusa/orchestration.py
+++ b/medusa/orchestration.py
@@ -65,7 +65,7 @@ class Orchestration(object):
                                 cert_file=cert_file)
             logging.debug('Batch #{i}: Running "{command}" on nodes {hosts} parallelism of {pool_size}'
                           .format(i=i, command=command, hosts=parallel_hosts, pool_size=len(parallel_hosts)))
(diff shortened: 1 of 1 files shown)


rzvoncek:
Hello @pstef , I've manged to carve out some time to look into this.
In principle, the change is great and we can accept it. But we need to introduce it more carefully. As done here, it'll change the old / default behaviour. That's not too great.
Instead, we'd need to hide this before a feature flag, meaning a new setting in the `[ssh]` section of the config file that would enable this.
Would you be willing to redo the change to do this? Or, would you be ok with me taking over?

pstef:
> Or, would you be ok with me taking over?

Please do :)

PR #799, opened 2024-08-14
Fix backup man future registration and BackupStatus computation of the node backup state

### LINKED ISSUE #797: Backup tracking is broken in the grpc server
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=74757756)

Backup tracking is failing to detect failed backups due to how the BackupStatus code does its check.

In the current version, the code will get the node backup from the storage bucket, and it will potentially register the backup in the BackupMan in case it doesn't exist. This allows to recreate the backups metadata in memory after a restart of the GRPC server.
It then computes the status solely based on the data gathered from storage, without checking if there is a pending future or not.

If there's no pending future (backup crashed or medusa was restarted), then we're not resurfacing the failure and the operator will keep on polling indefinitely.

What needs to be changed:
```
                BackupMan.register_backup(request.backupName, is_async=False, overwrite_existing=False)
                status = BackupMan.STATUS_UNKNOWN
                if backup.started:
                    status = BackupMan.STATUS_IN_PROGRESS
                if backup.finished:
                    status = BackupMan.STATUS_SUCCESS
```

this code needs to include a check for the existence of a future in the BackupMan and return accordingly:
```
                BackupMan.register_backup(request.backupName, is_async=False, overwrite_existing=False)
                status = BackupMan.STATUS_UNKNOWN
                try:
                   future = BackupMan.get_backup_future(request.backupName)
                except RuntimeError as e:
                  # No future exists, if the backup isn't marked as finished in the backend then it failed
                  if not backup.finished:
                    status = BackupMan.STATUS_FAILED
                if status == BackupMan.STATUS_UNKNOWN:
                  # We don't have a pending future and need to compute status based on storage information
                  if backup.started:
                      status = BackupMan.STATUS_IN_PROGRESS
                  if backup.finished:
                      status = BackupMan.STATUS_SUCCESS
```

adejanovski:
Fixes #797 

Two bugs are fixed in this PR:

- The `BackupMan.register_backup()` function was always registering the backup, even if it existed and overwrite was set to false due to bad code structure.
- The state was computed without taking into account the presence of a future. In case of a restart of the grpc server with an unfinished backup job, the backup would forever be seen as in progress even though it failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,12 @@ jobs:
         then
             echo "Updating debian changelog..."
             cd packaging/docker-build
-            docker-compose build release && docker-compose run release
+            docker compose build release && docker compose run release
             cd ../..
         fi
         cd packaging/docker-build
-        docker-compose build "cassandra-medusa-builder-${{ matrix.suite }}" \
-            && docker-compose run "cassandra-medusa-builder-${{ matrix.suite }}"
+        docker compose build "cassandra-medusa-builder-${{ matrix.suite }}" \
+            && docker compose run "cassandra-medusa-builder-${{ matrix.suite }}"
         
         cd ../../packages
         if [ -f "cassandra-medusa_${version}-0~${{ matrix.suite }}0_amd64.deb" ]; then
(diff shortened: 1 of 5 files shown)


adejanovski:
@rzvoncek, this is ready for review. I added some tests to check the behavior and manually checked that it works as expected in K8ssandra.
There are 2 failing tests which seem unrelated but I'll leave that up to you to decide (one seems to be a gcs communication issue and the other a webhook issue).

rzvoncek (approved):

rzvoncek:
I've retriggered the tests and they both passed.

PR #800, opened 2024-08-14
Do not set storage class for GCP (it's not supported anyway)

### LINKED ISSUE #798: GCP backups are failing if the storage class is unspecified
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=74758780)

We're missing a null check in the google storage implementation which makes the backups fail if the storage class isn't explicitly specified.

We instead should disable passing the storage class altogether since it's not supported by our gcp library underneath anyway.

In the Azure implementation, we need to add a null check and use the default storage class if it's not overridden.

### Definition of Done
- [ ] Storage class isn't used anymore in the google storage implementation
- [ ] A null check is added to the Azure storage implementation before calling `.upper()` on it

rzvoncek:
Fixes #798

### DIFF
--- a/medusa/storage/google_storage.py
+++ b/medusa/storage/google_storage.py
@@ -127,18 +127,15 @@ class GoogleStorage(AbstractStorage):
             )
         )
 
-        storage_class = self.get_storage_class()
-        ex_header = {"storageClass": storage_class} if storage_class else {}
         resp = await self.gcs_storage.upload(
             bucket=self.bucket_name,
             object_name=object_key,
             file_data=data,
             force_resumable_upload=True,
             timeout=-1,
-            headers=ex_header,
         )
         return AbstractBlob(
-            resp['name'], int(resp['size']), resp['md5Hash'], resp['timeCreated'], storage_class.upper()
+            resp['name'], int(resp['size']), resp['md5Hash'], resp['timeCreated'], None
         )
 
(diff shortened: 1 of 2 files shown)


medusa/storage/google_storage.py
```diff
@@ -127,18 +127,15 @@ async def _upload_object(self, data: io.BytesIO, object_key: str, headers: t.Dic
...
             bucket=self.bucket_name,
             object_name=object_key,
             file_data=data,
             force_resumable_upload=True,
             timeout=-1,
-            headers=ex_header,
         )
         return AbstractBlob(
-            resp['name'], int(resp['size']), resp['md5Hash'], resp['timeCreated'], storage_class.upper()
+            resp['name'], int(resp['size']), resp['md5Hash'], resp['timeCreated'], None
```
rzvoncek, line 138: In the other two providers, we use the information returned by the upload operation. In GCS case, we don't get this info, so we're not setting it at all.

adejanovski (approved):
Changes look good. Didn't we also have to patch the Azure storage class to add a null check?

rzvoncek:
Nope. The Azure class already does this:
```python
standard_blob_tier=StandardBlobTier(storage_class.capitalize()) if storage_class else None,
```

Issue #805, opened 2024-09-18
Cassandra pod stuck instead of CrashLoopBackOff when Medusa config loading fails


c3-clement:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=80216084)

See k8ssandra-operator issue: https://github.com/k8ssandra/k8ssandra-operator/issues/1406

**What happened?**

I deployed a K8ssandraCluster with 96 replicas and medusa enabled, and one of the pods did not reach the Readiness probe

```
k get sts cs-95f5cdf50d-cs-95f5cdf50d-default-sts -n platform
NAME                                      READY   AGE
cs-95f5cdf50d-cs-95f5cdf50d-default-sts   95/96
```

I identified the faulty pod: It was not reaching readiness probe because of the `medusa` container.
The `medusa` gRPC server did not start because `load_config()` failed (see logs below).
Since the gRPC server was not started, the readiness probe was not reached.

The `medusa` container was "blocked" and did not attempt to restart the gRPC server.
I restarted the pod manually by deleting it, and the medusa gRPC server started successfully.

**Did you expect to see something different?**

I expect the pod to restart and to be in `CrashLoopBackOff` phase if a uncaught exception is raised by the medusa python process, instead of blocking indefinitely.

I believe this behavior was introduced by the following change : https://github.com/thelastpickle/cassandra-medusa/pull/731

**How to reproduce it (as minimally and precisely as possible)**:
Start the medusa container with an invalid configuration

**Environment**

* K8ssandra Operator version:
1.18
* Medusa version:
0.21
* Kubernetes version information:
1.29
* Kubernetes cluster kind:
GKE

**Medusa logs**

```
MEDUSA_MODE = GRPC
sleeping for 0 sec
Starting Medusa gRPC service
WARNING:root:The CQL_USERNAME environment variable is deprecated and has been replaced by the MEDUSA_CQL_USERNAME variable
WARNING:root:The CQL_PASSWORD environment variable is deprecated and has been replaced by the MEDUSA_CQL_PASSWORD variable
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/cassandra/.venv/lib/python3.11/site-packages/medusa/service/grpc/server.py", line 424, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
[...]

PR #806, opened 2024-09-18
Propagate Medusa process exit code in k8s docker-entrypoint

c3-clement:
Resolve https://github.com/thelastpickle/cassandra-medusa/issues/805

Prior to this change, if the Medusa process exits with a non-zero exit code for some reason, the Medusa container will be stuck forever because of the `while true` loop.

With this fix, the while loop will be interrupted if the Medusa process exits, and the container will exit with the same code as the Medusa process exit code.

To validate this change, I've manually tested the following scenarios:
- [X] Start the medusa container without any file at `/etc/medusa.ini` -> the container should exit with a non-zero code.
- [X] Start the medusa container with a valid config at `/etc/medusa.ini` -> the container should run as expected.
- [X] Change the config `/etc/medusa.ini` while the container is running -> the container should exit with a zero code.
- [X] Put an arbitrary `sys.exit(42)` in `server.py` to mock a production issue -> the container should exit with the status code 42.

### DIFF
--- a/k8s/docker-entrypoint.sh
+++ b/k8s/docker-entrypoint.sh
@@ -11,12 +11,12 @@ set -e
 echo "MEDUSA_MODE = $MEDUSA_MODE"
 
 restore() {
-    # The BACKUP_NAME and RESTORE_KEY env vars have to be set in order for a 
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
LGTM, thanks for the fix @c3-clement !

PR #809, opened 2024-10-03
Radovan/tests for 529

rzvoncek:


.github/workflows/ci.yml
```diff
@@ -60,9 +60,9 @@ jobs:
           - suite: focal
             os-version: ubuntu-20.04
           - suite: bullseye
-            os-version: ubuntu-22.04
+            os-version: ubuntu-20.04
           - suite: bookworm
-            os-version: ubuntu-22.04
+            os-version: ubuntu-20.04
```
rtib, line 65: I've already tried that, it won't work. Ubuntu and Debian don't share their python infrastructure.

PR #813, opened 2024-10-09
MED-100 Make read_timeout default to None, only use it if set

### LINKED ISSUE #814: The `read_timeout` should default to nothing instead of 60 seconds
As we got notified [here](https://github.com/k8ssandra/k8ssandra-operator/issues/1353#issuecomment-2399309354), the way we introduced the read_timeout setting is not too compatible with existing instalations. The 60 seconds timeout is not good enough for everybody. 

We should default to no timeout, and only pass it in if it's actually configured.

rzvoncek:
Fixes #814

### DIFF
--- a/medusa/storage/google_storage.py
+++ b/medusa/storage/google_storage.py
@@ -49,7 +49,7 @@ class GoogleStorage(AbstractStorage):
 
         logging.getLogger('gcloud.aio.storage.storage').setLevel(logging.WARNING)
 
-        self.read_timeout = int(config.read_timeout)
+        self.read_timeout = int(config.read_timeout) if 'read_timeout' in dir(config) and config.read_timeout else None
 
         super().__init__(config)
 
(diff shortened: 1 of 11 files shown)


medusa/storage/google_storage.py
```diff
@@ -158,7 +158,7 @@ async def _download_blob(self, src: str, dest: str):
             stream = await self.gcs_storage.download_stream(
                 bucket=self.bucket_name,
                 object_name=object_key,
-                timeout=self.read_timeout if self.read_timeout is not None else -1,
+                timeout=self.read_timeout,
```
adejanovski, line 161: **Question:** Disabling the timeout seemed to be done by setting the value to `-1` previously and now it's set to `None`. Does that work as well?

rzvoncek: I actually do not know.
Since it's really tedious to find this out empirically, let's just default to -1 for GCS. The commit https://github.com/thelastpickle/cassandra-medusa/pull/813/commits/3dd0a55431e402436bb693e75748122e9313ddb9 does that.

adejanovski (approved):
Thanks for the efforts on fixing the tests!

PR #827, opened 2024-12-02
Add support for using an attached GCE service account

andyfoston:
Support using service accounts that are attached to a GCE instance, rather than requiring a service account key to be present.

The `gcloud-aio` library supports using a service account that is assigned to a VM by default just by omitting the `service_file` parameter (or by setting it to `None`).

Removing this requirement means that the service account key does not need to be manually rotated, as the assigned service account credentials provided by the GCE metadata API are automatically rotated frequently. 

Closes: #833

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -71,6 +71,7 @@ bucket_name = cassandra_backups
 kms_id = <ARN of KMS key used for server-side bucket encryption>
 
 ; JSON key file for service account with access to GCS bucket or AWS credentials file (home-dir/.aws/credentials)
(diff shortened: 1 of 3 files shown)


adejanovski:
Thanks for this PR @andyfoston, that's definitely something we want to support.
We'll circle back shortly with a review.

rzvoncek (approved):
Hi. I've tested this manually, it works like a charm.
The failing [integration-tests (3.9, local, 2.2.19, 8.0.252)](https://github.com/thelastpickle/cassandra-medusa/actions/runs/12121672569/job/33793215018?pr=827#logs) are a know issue, I'm fixing them elsewhere.

Issue #843, opened 2025-01-07
Fix the build build process for Medusa


andyfoston:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=92808436)

Medusa builds are currently failing with the below error:

```
Processing /Users/aft04/git/opensource/cassandra-medusa
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [14 lines of output]
      Traceback (most recent call last):
        File "/Users/aft04/venvs/cassandra-medusa-test/lib/python3.9/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/Users/aft04/venvs/cassandra-medusa-test/lib/python3.9/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/Users/aft04/venvs/cassandra-medusa-test/lib/python3.9/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 149, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
        File "/private/var/folders/gq/v5j4qzs11fqcjkfsztqxhm200000gp/T/pip-build-env-11iwth5i/overlay/lib/python3.9/site-packages/poetry/core/masonry/api.py", line 42, in prepare_metadata_for_build_wheel
          poetry = Factory().create_poetry(Path().resolve(), with_groups=False)
        File "/private/var/folders/gq/v5j4qzs11fqcjkfsztqxhm200000gp/T/pip-build-env-11iwth5i/overlay/lib/python3.9/site-packages/poetry/core/factory.py", line 58, in create_poetry
          raise RuntimeError("The Poetry configuration is invalid:\n" + message)
      RuntimeError: The Poetry configuration is invalid:
        - project must contain ['name'] properties

      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

This happens for all version from the git tag v0.20.0 until HEAD

[...]

PR #844, opened 2025-01-07
Update poetry-core version to be <2.0.0

andyfoston:
The build process for medusa is currently failing due to a mis-configuration with Poetry. It appears that the latest version of Poetry has introduced breaking changes.

This change prevents version 2.0.0 upwards from being used to ensure build continue to work

fixes: #843


rzvoncek (approved):

rzvoncek:
Thanks a lot @andyfoston !

PR #845, opened 2025-01-08
Radovan/529 take2

rzvoncek:


rtib:
have tested bullseye and bookworm builds
LGTM :+1:

PR #852, opened 2025-02-06
851 Backup Without Snapshot Creation & Deletion

### LINKED ISSUE #851: Backup Without Snapshot Creation & Deletion
### Description

It would be useful to have an option to run the backup feature in Medusa without automatically creating a snapshot. Instead, I would like to use Medusa solely for uploading an existing snapshot to the backup repository. Additionally, I would prefer Medusa not to delete the snapshot after the backup process is completed.

### Use Case

I have my own workflow for creating snapshots, and I want to leverage Medusa only for the upload process. Currently, it seems Medusa always takes a new snapshot and deletes it afterward, which doesn’t align with my needs.

### Possible Solution

Would it be possible to introduce an option (e.g., --no-snapshot) that allows users to specify an existing snapshot for upload instead of creating one? Additionally, a flag like --keep-snapshot could be useful to prevent automatic deletion.

### Notes

I’m not an expert in Medusa, so I might be misunderstanding how it works. If there’s already a way to achieve this, I’d appreciate any guidance.

Thanks for your time and for maintaining this project!

PACordonnier:
Fixes #851 

Hi,

Added two flags to the "backup" command. The default behaviour of backup is unchanged : it creates a snapshot and deletes it on successful backup.

New flags:

--keep-snapshot: dont delete snapshot on successful backup.
--use-existing-snapshot: dont create a snapshot, use a previously existing one.

- Raises an error when --use-existing-snapshot is set and --backup-name is not defined.
- Raises an error when --use-existing-snapshot and snapshot cant be found.

If the feature is accepted in principle, I’m happy to explore ways to add tests. Currently the integration_steps does not manage nodetool operations needed for theses features (such as `snapshot` and `listsnapshots`).

Also I can work on adding feature to the backup-cluster. Also is more code needed for DSE environment ?

Looking forward to your thoughts—thanks for your time!

### DIFF
--- a/medusa/backup_node.py
+++ b/medusa/backup_node.py
@@ -75,7 +75,8 @@ def stagger(fqdn, storage, tokenmap):
 # Called by async thread for backup, called in main thread as synchronous backup.
 # Kicks off the node backup unit of work and registers for backup queries.
 # No return value for async mode, throws back exception for failed kickoff.
-def handle_backup(config, backup_name_arg, stagger_time, enable_md5_checks_flag, mode, keep_snapshot=False):
+def handle_backup(config, backup_name_arg, stagger_time, enable_md5_checks_flag, mode, keep_snapshot=False,
+                  use_existing_snapshot=False):
     start = datetime.datetime.now()
     backup_name = backup_name_arg or start.strftime('%Y%m%d%H%M')
     monitoring = Monitoring(config=config.monitoring)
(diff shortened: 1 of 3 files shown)


rzvoncek:
Hello @PACordonnier ! I finally managed to have a look at the change. I generally like it - it's fairly minimalistic, and seems to honour our principle of backwards compatibility.

However, we cannot quite accept it without having some tests. Ideally, we'd add an integration test scenario that would cover all the variants of this. Is that something you could tackle too?

PACordonnier:
Hey @rzvoncek thanks a lot for your feedback, much appreciated.

I will work on adding some tests, I'm currently on holiday but will check it soon as I get back 👍

PACordonnier:
Hi @rzvoncek 

I've added two integration tests. CI is not passing but its not relatedt to the new tests.

I could not add tests for testing if code raises issues.

 - The error `Cannot use existing snapshot without providing a backup name` is raised in medusacli, which is not tested
 - When the error `Snapshot {} does not exist and use_existing_snapshot is True. Please create the snapshot first` is raised, the way medusa handles errors ends up on a Sys.exit(1) and are not thrown back to the backup.

Please tell me if you require more tests or change I will gladly work on it. Thanks for your time

rzvoncek:
Hello @PACordonnier! Big thanks for pushing things. I'll try to have a look in the upcoming days.

rzvoncek:
Hi @PACordonnier , I'd just like to keep you in the loop - I'm swamped and not managing to look into this just yet. I'll keep you posted once I get into this.

rzvoncek:
Hello, I've finally managed to get some time to look into this. Overall, the change/feature looks great. There are just 3 small(ish) things I think we should improvie:
* Add the use/keep snapshot options to `backup-cluster` command as well.
* Adjust or add a scenario to actually use an older (not the most recent) snapshot for backup.
* Fix the ccm list snapshots command in ITs. There are more than 2 "garbage" log lines in some environments (the CI and my laptop).

I've done all these 3 things over [here](https://github.com/thelastpickle/cassandra-medusa/pull/871). If you'll manage to have a look, that's awesome. Otherwise, I'm gonna merge this on Friday (tomorrow).

PACordonnier:
That's awesome, really appreciate. Everything looks good.

Thanks a lot @rzvoncek

PR #865, opened 2025-06-09
Build images off Ubuntu 24.04 + few python dep updates

rzvoncek:


k8s/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=linux/${TARGETARCH} ubuntu:22.04 AS base
+FROM --platform=linux/${TARGETARCH} ubuntu:24.04 AS base
```
adejanovski, line 1: **thought:** we might want to move to much leaner images such as the python alpine ones at some point, or maybe a slim ubuntu if that's applicable.
24.04 still reports some CVES:
Capture d’écran 2025-06-13 à 18 28 48

3.12-alpine would work CVE wise: https://hub.docker.com/layers/library/python/3.12-alpine3.22/images/sha256-c7faa49938b528b23f0d3f900d6c3e11e16ded420852d327f58267547d26f915

PR #885, opened 2025-08-05
Replace retrying with tenacity for @retry

idleyoungman:
The maintainer says retrying has no support for async functions: https://github.com/groodt/retrying/issues/8#issuecomment-2993942195

But `@retry` is used on several async functions in medusa.

The tenacity fork of retrying does have async support: https://github.com/jd/tenacity?tab=readme-ov-file#async-and-retry

`tenacity` uses seconds for unformatted time values, rather than ms.


idleyoungman:
Looks like replacing retrying with tenacity happened in #896. Closing.

PR #907, opened 2025-11-14
Add configuration option for gRPC interface

andy-slac:
I am using Medusa gRPC service in a non-conventional setup without kubernetes. It works fine, but there is one minor issue - by default it binds itself to `[::]` interface which is wide open to the whole network. In my case it would be better to limit it to a loopback interface (127.0.0.1), but that is not possible as the interface is hardcoded. This patch adds an option to specify interface in a configuration file.

All tests pass, but I did not add a unit test to tests non-default interface. I could potentially add a case with `interface` set to something like `127.0.0.2` if this is desired.

### DIFF
--- a/medusa-example.ini
+++ b/medusa-example.ini
@@ -186,6 +186,9 @@ use_sudo_for_restore = True
 ; Allows to propagate the exceptions instead of exiting the program.
 ;enabled = False
 
+; Interface that service will listen.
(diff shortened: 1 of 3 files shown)


andy-slac:
Linked issue #964

andy-slac:
One check fails, but I think it fails in other PRs as well.

PR #917, opened 2025-12-22
feat: Implement concurrent_transfers for GCS storage

andyfoston:
Implement the `concurrent_transfers` feature for GCS storage. This re-uses the same parameter that is implemented for other storage backends.

This is implemented using `asyncio.Semaphore` to restrict the number of transfers happening concurrently


andyfoston:
Hi, please can I get some feedback on this, as we are still seeing issues in our GCP environment that I believe this PR would solve. Many thanks

rzvoncek:
Hi, tested this, seems to work fine! Thanks for the effort, and sorry for the wait.

Issue #892, opened 2025-09-08
Add client-side encryption for backup files


fasar:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=127838371)

It would be highly valuable (for me) to support client-side encryption for backup files in Medusa.
I would like to encrypt data on the Medusa before uploading it to the storage backend (S3, Azure, GCS, etc.).
This would improve security by ensuring that backups are protected and not depends of the backend's encryption capabilities.

If anyone else in the community is interested in this feature, please feel free to comment or share your use case.
I would be glad to collaborate or help coordinate efforts if others want to contribute or provide feedback.

In a first approach, I would create a wrapper based on [abstract_storage.py](https://github.com/thelastpickle/cassandra-medusa/blob/master/medusa/storage/abstract_storage.py) and implement a hook for the `_upload_blob` and `_download_blob` functions. Also, I would use the Fernet library for encryption and decryption.

I would like to know where to start:

- Guidance on where to start in the codebase to implement if not in a wrapper of abstract_storage.
- Recommendations for how to structure the feature (CLI options, configuration file changes).
- Advice on the current and commonly used options available for cryptography software.
- Advice on which unit/integration tests should be implemented or updated to ensure robust coverage for this feature.
- Any caveats or best practices to follow in the context of Medusa (for example: update the documetation ...).

Thank you in advance,

fasar:
Hi community,

If I develop this feature and open a pull request, would you be open to reviewing it for possible inclusion in the main branch?
I’d like to make sure this aligns with the project’s goals before I start working on it.

Thank you in advance,
Fabien.

alsarins:
I strongly support this initiative. Take my vote) This would be very useful, as server-side encryption isn't an option in case when database contains personal or financial data. 
Client-side encryption is really good choice here.

We use s3 compatible storages (ceph and minio) for all backups of our various database types, but as far as databases keep personal data, we do not trust s3 storage owners, and sending backups already encrypted at client side is mandatory for us. 

[...]

PR #928, opened 2026-01-23
Fsa clientside encryption

fasar:
This PR implements client-side encryption for backup files in Medusa, addressing issue #892.

This feature allows users to encrypt data before uploading to storage backends (S3, Azure, GCS, MinIO, IBM Cloud Object Storage), ensuring that backups are protected independently of the backend's encryption capabilities.

I didn't add any dependencies. I was helped a lot by AI tools. :-D
This implementation uses a temporary directory to encrypt files on the local disk before sending them.

This could be improved by using the [AWS Encryption SDK](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/python-example-code.html#python-example-streams), but my Python skills are not sufficient to handle encryption and MD5 calculation on the fly.
Additionally, adding the AWS Encryption SDK would require updating the project dependencies.


tobast:
Thanks for your work! I was just now searching whether there were plans to add encryption support for Medusa.

I am however a bit concerned by your description of this merge request. Was a substantial part of it written by LLMs? I am not really at ease with the idea of implementing encryption through AI, even when the implementation of cryptography itself is from a third-party library.

fasar:
Hello Tobast,

Thank you for your comment.
I can assure you that I read all the production code line by line.
I spent two weeks finalizing and tuning all the code I added.

I'm a beginner in Python, so if you have expertise, we can connect and review the code together :-D

Thanks,
Fabien.

PR #954, opened 2026-04-22
Add AWS CRT-based S3 transfer support via s3transfer.crt

### LINKED ISSUE #870: CRT Transfer Manager
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=115715075)

I am running the latest version of medusa on my EKS Cluster and I see this message in the logs `DEBUG:boto3.s3.transfer:Opting out of CRT Transfer Manager. Preferred client: auto, CRT available: False, Instance Optimized: False.`
AWS Documentation states that there is significant improvement in using  [CRT](https://aws.amazon.com/blogs/developer/introducing-crt-based-s3-client-and-the-s3-transfer-manager-in-the-aws-sdk-for-java-2-x/). I would like thoughts of the community in incorporating  it in the code base

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MED-134) by [Unito](https://www.unito.io)
┆Issue Number: MED-134

SreedevT:
Closes #870

## Summary
Introduces a new `use_crt` config option that enables the AWS Common Runtime (CRT) transfer client for S3 uploads and downloads, bypassing boto3's instance-type gating by using s3transfer.crt.CRTTransferManager directly.

- Add `use_crt` to StorageConfig namedtuple with default False
- In connect(), build a CRTTransferManager when use_crt=True using create_s3_crt_client, BotocoreCRTCredentialsWrapper, and BotocoreCRTRequestSerializer; acquire the CRT process lock to prevent multiple managers in the same process
- Fall back gracefully if storage_provider is s3_compatible (CRT does not support custom endpoints)
- Use CRTTransferManager.download()/upload() with file objects in __download_blob and __upload_file when CRT is active
- Shut down CRT manager cleanly in disconnect()
- Add awscrt==0.32.1 as a default dependency
- Document use_crt in medusa-example.ini
- Add unit tests covering default and enabled CRT cases

## Caveats
use_crt is only supported when medusa runs via the gRPC server entry point. The medusa CLI calls gevent.monkey.patch_all() at startup, which replaces threading primitives with libev-backed versions. CRT's native C threads signal completion via pthread cond and gevent never sees the signal and CRT futures deadlock forever. load_config() detects this and refuses to start with use_crt=True under the CLI, exiting with a clear error.


SreedevT:
@rzvoncek Could you review this

SreedevT:
I was able to test out restore on a cluster with around 800GB data. It was around 2.5x faster than default values on medusa and even with tuned values it was 10% faster. There are a few caveats though, CRT saturates the network and there is no way to control the throughput as of now [ref](https://github.com/aws/aws-cli/issues/8974). This is not ideal on the backup path where the network of a live cluster will get saturated. I do think its still useful for the restore path where network saturation is fine.

medusa/config.py
```diff
@@ -374,6 +375,16 @@ def load_config(args, config_file):
             logging.error('Required configuration "{}" cannot contain a slash ("/")'.format(field))
             sys.exit(2)
 
+    if evaluate_boolean(getattr(medusa_config.storage, 'use_crt', 'False')):
```
SreedevT, line 378: check and exit on config if using medusa cli and crt are enabled together

pyproject.toml
```diff
@@ -82,6 +82,7 @@ azure-storage-blob = "12.17.0"
 urllib3 = "2.6.3"
 pyjwt = "2.12.0"
 pyasn1 = "0.6.3"
+awscrt = "0.32.1"
```
SreedevT, line 85: This is a hard dependency now.

medusa/storage/s3_base_storage.py
```diff
@@ -162,9 +178,67 @@ def connect(self):
...
+        resolved_credentials = self.s3_client._get_credentials()
+        crt_credentials = BotocoreCRTCredentialsWrapper(resolved_credentials)
+        crt_credentials_provider = crt_credentials.to_crt_credentials_provider()
+
+        crt_client_kwargs = {
+            'region': self.credentials.region,
+            'crt_credentials_provider': crt_credentials_provider,
+        }
+
+        multipart_chunksize = self.config.multipart_chunksize or None
```
SreedevT, line 204: Some configs dont have parity across boto3 and crt. We could either skip or add new variables that are CRT specific. Skipped for now but open for change

SreedevT:
I also noticed that there are no integration tests for aws/s3, so this feature is not being checked on ci

PR #968, opened 2026-06-26
Fix GetBackups error response status field

### LINKED ISSUE #818: grpc error
https://github.com/thelastpickle/cassandra-medusa/blob/1394391ca9bbb9477137d989c34199cb215b6bd1/medusa/service/grpc/server.py#L265

I got error when the get_backups failed (due to azure blog credential config error), then the exception part also throw exception because the `GetBackupsResponse` doesn't have `status` field. Is this a bug, I think it should be something like `response.overallStatus = ?` right?

```
message GetBackupsResponse {
  repeated BackupSummary backups = 1;
  StatusType overallStatus = 2;
}
```

```
    def GetBackups(self, request, context):
        response = medusa_pb2.GetBackupsResponse()
        try:
            # cluster backups
            with Storage(config=self.storage_config) as connected_storage:
                backups = get_backups(connected_storage, self.config, True)
                for backup in backups:
                    summary = get_backup_summary(backup)
                    response.backups.append(summary)
                set_overall_status(response)

        except Exception as e:
            context.set_details("Failed to get backups due to error: {}".format(e))
            context.set_code(grpc.StatusCode.INTERNAL)
            response.status = medusa_pb2.StatusType.UNKNOWN
        return response
```

fuqi1001:
## Summary

Fixes #818.

This updates the `GetBackups` error path to set `overallStatus` instead of `status` on `GetBackupsResponse`. The protobuf response message does not define a `status` field, so the previous error handling could raise an `AttributeError` while trying to handle the original failure.

## Changes

- Set `response.overallStatus` to `UNKNOWN` when `GetBackups` fails.
- Add a regression test for the `GetBackups` failure path.

## Testing

- `compile()` syntax check for `medusa/service/grpc/server.py` and `tests/service/grpc/server_test.py`
- `git diff --check`

### DIFF
--- a/medusa/service/grpc/server.py
+++ b/medusa/service/grpc/server.py
@@ -273,7 +273,7 @@ class MedusaService(medusa_pb2_grpc.MedusaServicer):
         except Exception as e:
             context.set_details("Failed to get backups due to error: {}".format(e))
             context.set_code(grpc.StatusCode.INTERNAL)
-            response.status = medusa_pb2.StatusType.UNKNOWN
+            response.overallStatus = medusa_pb2.StatusType.UNKNOWN
         return response
 
     def DeleteBackup(self, request, context):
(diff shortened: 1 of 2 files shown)


rzvoncek (approved):
Thanks @fuqi1001 !

Issue #754, opened 2024-04-26
Differential mode re-uploading all files from host when --prefix is not added


sunyoona:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=61063762)

Hello there,

- I find that when I use medusa --prefix 111 backup --backup-name test1 --mode differential, for each differential backup, some files can be copied from the previous backup.

First differential backup:

`$ nohup medusa --prefix 2024-04-26 backup --backup-name 111 --mode differential > ~/backup/111.log 2>&1 &

[2024-04-26 03:36:04,030] INFO: Updating backup index
[2024-04-26 03:36:04,330] INFO: Backup done
[2024-04-26 03:36:04,330] INFO: - Started: 2024-04-26 03:30:19
                        - Started extracting data: 2024-04-26 03:30:20
                        - Finished: 2024-04-26 03:36:04
[2024-04-26 03:36:04,331] INFO: - Real duration: 0:05:43.582589 (excludes time waiting for other nodes)
[2024-04-26 03:36:04,332] INFO: - 616 files, 9.64 GB
[2024-04-26 03:36:04,332] INFO: - 616 files copied from host (616 new, 0 reuploaded)
[2024-04-26 03:36:04,332] INFO: - 0 kept from previous backup (111)
`

Second differential backup:
`$ nohup medusa --prefix 2024-04-26 backup --backup-name 222 --mode differential > ~/backup/222.log 2>&1 &

[2024-04-26 03:37:01,692] INFO: - Started: 2024-04-26 03:36:58
                        - Started extracting data: 2024-04-26 03:36:59
                        - Finished: 2024-04-26 03:37:01
[2024-04-26 03:37:01,692] INFO: - Real duration: 0:00:02.405600 (excludes time waiting for other nodes)
[2024-04-26 03:37:01,693] INFO: - 600 files, 9.64 GB
[2024-04-26 03:37:01,693] INFO: - 48 files copied from host (48 new, 0 reuploaded)
[2024-04-26 03:37:01,693] INFO: - 552 kept from previous backup (222)`

But when I use medusa backup --backup-name test2 --mode differential, for every differential backup, the files are all reuploaded from the host.

First differential backup:
`$ nohup medusa  backup --backup-name 111 --mode differential > ~/backup/111.log 2>&1 &

[2024-04-26 03:37:34,154] INFO: Updating backup index
[2024-04-26 03:37:34,581] INFO: Backup done
[...]

PR #969, opened 2026-07-02
Fix S3 upload regressions from the libcloud→boto3 transition (connection pool, threshold mismatch, chunk size, parallelism, S3-compatible fix)

arodrime:
## Problem

S3 backups were taking much longer than expected (roughly from 4h to 24h over successive Medusa versions). Started after upgrading from 0.14 to 0.28 — likely the trigger. Logs flooded with:

```
WARNING: Connection pool is full, discarding connection: <s3-endpoint>. Connection pool size: 8
```

Raising `concurrent_transfers` didn't help. This isn't just cosmetic: every discarded connection means urllib3 drops and re-establishes a fresh TCP+TLS handshake instead of reusing a pooled one, which adds up fast across a backup with thousands of SSTable files. On shared S3 endpoints, the resulting flood of redundant requests plus needless reuploads may also be degrading things for other tenants on the same storage. This is still unconfirmed, isolated field tests are planned on upcoming days.

Separately, differential backups were systematically reuploading files already safely in storage — but only when `enable_md5_checks=True` is set (not the default). Root cause: `multi_part_upload_threshold` was never passed to boto3, so boto3 always used its own 8MB default to decide whether a file gets multipart-uploaded (and therefore which ETag format S3 stores), while Medusa used the *configured* (higher) threshold to decide which hash algorithm to verify with. With a threshold of e.g. 100MB, every file between 8MB and 100MB is multipart from boto3's point of view but single-part from Medusa's. The hashes can never match, so those files get reuploaded on every single differential backup regardless of whether they changed — as long as md5 checks are enabled. With md5 checks off (default), this specific bug has no effect: verification falls back to size comparison alone, and
`threshold` is never even read. 
Before/after from a real backup (params changed together, so I can't yet attribute the improvement to one fix specifically — see test plan):

```
# concurrent_transfers=4, multi_part_upload_threshold=200MB (unpatched)
- 9800 files, 380.42 GB
- 2560 files copied from host (1620 new, 940 reuploaded)
- Real duration: 9:12:44
WARNING: Connection pool is full, discarding connection...

- 9950 files, 381.17 GB
- 2340 files copied from host (1890 new, 450 reuploaded)
- Real duration: 14:38:52
WARNING: Connection pool is full, discarding connection...

# concurrent_transfers=1, multi_part_upload_threshold=8MB (workaround)
- 9950 files, 381.17 GB
- 530 files copied from host (530 new, 0 reuploaded)
- Real duration: 0:21:07
```

My guess is both bugs contribute independently: pool saturation causes verification hiccups regardless of file size, while the threshold mismatch causes systematic reuploads in the 8MB-threshold band regardless of pool state. This PR fixes both, plus a third related bug (multipart chunk size hardcoded on one side, configurable on the other).

Same symptoms have been reported before without a root cause identified:
- #734 - `concurrent_transfers` raised from 4 to 16 to fight this warning, without success
- #754 - same warning reported alongside differential-mode reupload issues

## Issues and fixes

**1. Connection pool undersized** (ce6fe90)

Pool was sized as `concurrent_transfers * 2`, but boto3's `max_concurrency` was separately hard-coded to `4` threads per transfer. At `concurrent_transfers=4`, actual need (16) already exceeded the pool (8) — and raising `concurrent_transfers` scales both sides by the same ratio, so it never catches up. Matches #734 exactly. Fixed by making `max_concurrency` configurable (`multipart_max_concurrency`) and sizing the pool as `concurrent_transfers * multipart_max_concurrency + 10` (the `+10` is headroom for other incidental pool usage).

**2. `multi_part_upload_threshold` silently ignored by boto3** (baf0e81)

Never passed to boto3's `TransferConfig`, so boto3 used its own 8MB default no matter what was configured, while Medusa's own verification logic used the real configured value. Files in between get reuploaded each time, reducing the benefit of the differential backup. Fixed by passing the threshold through to boto3.

**3. MD5 verification chunk size didn't match the real upload** (6f94e60)

Same class of bug as the on one above: the local multipart-hash recomputation used a hard-coded 8MB part size, completely independent of the `multipart_chunksize` actually used by boto3. Once chunk size differs from 8MB, verification breaks. Making it a parameter fixes that.

**4. `multi_part_upload_threshold` now accepts human-readable sizes** (c1e20d6)

e.g. `20MB` instead of `20971520`, consistent with `multipart_chunksize`/`transfer_max_bandwidth`. Old raw byte values still work.

**5. Fixed-size batches replaced with a sliding window** (c498fc8)

_upload_blobs split the file list into fixed-size chunks of concurrent_transfers and awaited each chunk fully (asyncio.gather per chunk) before starting the next one. A single slow, large, or retried file anywhere in a chunk blocked every other file in that same chunk, and the next chunk wouldn't start until the current one fully drained, no matter how high concurrent_transfers was set.
Real example from a run with concurrent_transfers=4, files all get uploaded within the same second, then nothing for over a minute before the next sstable starts:

Logs look like this with concurrent_transfer set to 4:

```
16:03:38,445  Filter.db            232.000B
16:03:38,446  Data.db              162.060MiB
16:03:38,446  Index.db              28.453KiB
16:03:38,446  Summary.db           227.000B

# --- stall ~65s ---

16:04:43,845  CompressionInfo.db    57.874KiB
16:04:43,845  Statistics.db          5.524KiB
16:04:43,846  TOC.txt               92.000B
16:04:43,846  Digest.crc32          10.000B

# --- almost immediate, no big files above ---

16:04:44,205  Filter.db            168.000B
16:04:44,205  Data.db              163.216MiB
16:04:44,205  Index.db              29.074KiB
16:04:44,206  Summary.db           224.000B

# --- stall ~82s ---

16:06:06,988  CompressionInfo.db    66.093KiB
16:06:06,988  Statistics.db          5.193KiB
16:06:06,988  TOC.txt          92.000B
16:06:06,989  Digest.crc32          10.000B
```

**6. botocore's new default checksum behavior breaks some S3-compatible endpoints** (466c76c)

Since botocore 1.36, boto3 defaults to always computing/validating flexible checksums (CRC32/CRC64) on requests that support them, using chunked transfer encoding with a trailing checksum ("STREAMING-UNSIGNED-PAYLOAD-TRAILER"). Several S3-compatible implementations don't support this format and reject the request with `XAmzContentSHA256Mismatch`. I hit this against an internal S3-compatible endpoint (different from the S3-compatible we generally use) while testing this branch, on the very first `PutObject` call (the schema upload), unrelated to the pool/threshold/batching fixes above. Fixed by setting `request_checksum_calculation`/`response_checksum_validation` to `when_required` in the boto3 `Config`, restoring the pre-1.36 behavior of only computing/validating checksums
when the API operation actually requires them. Known upstream issue: [boto/boto3#3738](https://github.com/boto/boto3/issues/3738). 

## Tests added 
- `test_connect_sets_pool_size_from_config` --> pool sized as `concurrent_transfers * multipart_max_concurrency + 10`
- `test_transfer_config_uses_configured_values` --> `multi_part_upload_threshold`/`multipart_chunksize` actually reach boto3's `TransferConfig`
- `test_multi_part_s3_file_uses_configured_chunk_size` --> MD5 verification uses the configured chunk size
- `test_multi_part_s3_file_wrong_chunk_size_fails` --> MD5 verification fails when chunk size doesn't match the real upload
- `test_convert_human_friendly_size_to_bytes` / `test_convert_bare_byte_count_to_bytes` / `test_convert_invalid_size_raises` --> human-readable and legacy threshold formats
- `test_upload_blobs_uses_sliding_window_not_fixed_batches` (point 5) --> proves the batch-scheduling fix deterministically (fails on pre-fix code, passes after)
-  `test_connect_disables_flexible_checksums` (point 6) --> verifies `connect()` sets both `request_checksum_calculation` and `response_checksum_validation` to `when_required` on the boto3 `Config`

These are quite impactful bug fixes/improvements and should hopefully justify a release after testing and review.


rzvoncek:
Hi @arodrime . I've just merged another PR that re-enables S3 tests. Could you please rebase on a recent master to bring the newly added tests in?

arodrime:
> Hi @arodrime . I've just merged another PR that re-enables S3 tests. Could you please rebase on a recent master to bring the newly added tests in?

Done @rzvoncek, thanks for looking into this. I can rebase other branches as well as we move forward with master/main branch, just ping me as you did here when it makes sense to do it.

This PR is running in our production (using md5 checksums) and it reduced our backups from 24-36+ hours to 1-4 hours depending on the cluster. I think this PR is the most important of the few PRs I sent. It fixes flows/regressions in the design introduced a long time ago when we went from libcloud to boto3. In my current team, it became obvious as we moved all the way from medusa 0.14 to 0.29 and saw huge perfromances degradations. Yey, my environment is just one of many, so I am happy to have you check this.

arodrime:
Hi @rzvoncek, rebased onto master. CI is showing some failures, but they all look like infra flakes, not related to this PR:

- integration-tests (3.10, local, 4.0.18), (3.12, azure, 5.0.3), (3.12, local, 5.0.3): all fail at Install dependencies with github.com/riptano/ccm.git returning HTTP 503 during clone.
- k8ssandra-CreateSingleMedusaJob: fails at Login to DockerHub with "Username and password required" — looks like a missing/expired secret in the workflow.
- SonarCloud / Snyk code analysis: failed shortly after the force-push, likely a timing issue comparing against the new base.

I didn't look in details. Let me know if you'd like me to re-run the test (but I feel they always fail, not just flaky but kind of broken), wait for a fix, or anything else.

rzvoncek (approved):

rzvoncek:
Thanks for the rebase. I've re-ran the flakes, they are all green now.

I think I'll get to have a look at the other PRs within upcoming several days. So if you have a moment, rebasing them would be super helpful.

arodrime:
@rzvoncek one last thing about this PR. When I installed it in my client's infra, we literally destroyed our S3-compatible storage, we totally flooded it. I would add this to the changes.md if you agree, feel free to add it:


Heads up for upgraders: 

-----

* if you previously raised concurrent_transfers and/or multipart_max_concurrency to work around slow backups, those values were largely ineffective before this fix (pool was hard-capped independent of config). After this PR, they're fully honored; the real number of concurrent connections/threads will jump to match.
* If your S3-compatible endpoint has its own connection/rate limits (several do), you may see throttling, connection resets, or worse throughput than before right after upgrading. 

--> Recommend auditing concurrent_transfers/multipart_max_concurrency, or even transfer_max_bandwidth, back down to sane values before upgrading, then raising gradually while watching for pool/throttling warnings. 

----- 

In our case we had transfer_max_bandwidth=2048MB/s configured (really high considering it is per node!) as an attempt to speedup backups and it was never a real constraint before since actual throughput was capped well below that by the issues above, and it became an issue immediately after upgrading. I assume people will find out by themselves, but if we can warn people who actually take the time to read the changes, then I guess it is a win.

arodrime:
> Thanks for the rebase. I've re-ran the flakes, they are all green now.
> 
> I think I'll get to have a look at the other PRs within upcoming several days. So if you have a moment, rebasing them would be super helpful.

All PRs up to date and (re)based on master (one was built on top of #969).

Issue #155, opened 2020-06-03
Provide telemetry to report the progress of backups


jsanda:
I was looking over Instaclustr's backup tool which is integrated with their Kubernetes operator for Cassandra. It provides status like this:

```
$ kubectl describe cassandrabackups.cassandraoperator.instaclustr.com 
Name:             test-cassandra-backup-restore-s3
... other fields omitted
Status:
  Node:      cassandra-test-dc-cassandra-west1-a-0
  Progress:  66%
  State:     RUNNING
  Node:      cassandra-test-dc-cassandra-west1-b-0
  Progress:  48%
  State:     RUNNING
  Node:      cassandra-test-dc-cassandra-west1-c-0
  Progress:  45%
  State:     RUNNING
```
Reporting the progress of backups could be really useful for monitoring and for debugging. Maybe it could be as simple as reporting the percentage of bytes uploaded.

PR #972, opened 2026-07-15
Parallelize MD5 checksum comparison and add debug progress logging

arodrime:
## Summary

- With differential backups and `enable_md5_checks` on, comparing local files against the manifest was a fully sequential, silent loop (`check_already_uploaded`) - on large tables this could dominate total backup
time with zero visibility.
- Adds debug-level progress logging for this comparison pass (start + periodic progress).
- Parallelizes the comparison across a new dedicated `md5_check_concurrency` setting (`[checks]`, default `1`). Deliberately not reusing `concurrent_transfers`: that setting governs network upload concurrency,
while this is local CPU-bound hashing work - Medusa runs on the same node as Cassandra, so the default stays sequential and must be raised explicitly. Default behavior is sequential and not changed.

**Depends on #969** (not yet merged): this branch is stacked on `fix-s3-connection-pool-size` because `check_already_uploaded`/`file_matches_storage` already take `multipart_chunksize` there. The diff below
includes #969's commits until it merges; once it does, this diff will shrink to just the two commits above it. Hopefully.

Related to #155  - but not a full fix. This PR should ensure there is always something logged at the debug level, but is not tracking progress (% / Bytes etc).

## Tests
- Added `tests/backup_node_test.py::test_check_already_uploaded_respects_md5_check_concurrency` - verifies actual concurrency is capped at the configured value

### DIFF
--- a/medusa-example.ini
+++ b/medusa-example.ini
@@ -178,6 +178,14 @@ use_sudo_for_restore = True
 ;expected_result = <Coma separated string representation of values returned by the query. Checks only 1st row returned, and only if specified>
 ;enable_md5_checks = <During backups and verify, use md5 calculations to determine file integrity (in addition to size, which is used by default)>
 
+; When enable_md5_checks is on, number of files hashed in parallel while comparing local files
+; against the manifest in differential backups. This is local CPU-bound work (not network) - on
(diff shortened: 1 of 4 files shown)


rzvoncek (approved):
It worked, there are just 2 commints now. looks good.

PR #973, opened 2026-07-15
Don't crash the backup when a DSE node has no metadata folder & Read from cassandra.yaml

### LINKED ISSUE #819: cassandra-medusa installation issues on Amazon Linux 2 with python 3.8.16
Hi,

I was trying to setup latest version of Cassandra-medusa (0.22.3) on Amazon Linux2 but facing below dependency conflicts with python 3.8.16

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cassandra-medusa 0.22.3 requires botocore==1.31.75, but you have botocore 1.35.61 which is incompatible.
boto3 1.28.71 requires botocore<1.32.0,>=1.31.71, but you have botocore 1.35.61 which is incompatible.
boto3 1.28.71 requires s3transfer<0.8.0,>=0.7.0, but you have s3transfer 0.10.3 which is incompatible.
```

After installing medusa, it's also causing issues with awscli commands. One such error when I was trying to call `aws s3` after medusa installation. 

```
  File "/usr/local/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/local/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/usr/local/lib/python3.8/site-packages/awscli/clidriver.py", line 73, in main
    driver = create_clidriver()
  File "/usr/local/lib/python3.8/site-packages/awscli/clidriver.py", line 82, in create_clidriver
    load_plugins(
  File "/usr/local/lib/python3.8/site-packages/awscli/plugin.py", line 44, in load_plugins
    modules = _import_plugins(plugin_mapping)
  File "/usr/local/lib/python3.8/site-packages/awscli/plugin.py", line 61, in _import_plugins
    module = __import__(path, fromlist=[module])
  File "/usr/local/lib/python3.8/site-packages/awscli/handlers.py", line 106, in <module>
    from awscli.customizations.s3.s3 import s3_plugin_initialize
  File "/usr/local/lib/python3.8/site-packages/awscli/customizations/s3/s3.py", line 15, in <module>
    from awscli.customizations.s3.subcommands import ListCommand, WebsiteCommand, \
  File "/usr/local/lib/python3.8/site-packages/awscli/customizations/s3/subcommands.py", line 18, in <module>
    from botocore.utils import is_s3express_bucket, ensure_boolean
ImportError: cannot import name 'is_s3express_bucket' from 'botocore.utils' (/usr/local/lib/python3.8/site-packages/botocore/utils.py)
````

As highlighted in the documentation, even though I'm using python version 3.8 as a pre-req may I know am I hitting above errors? I'm I missing anything? 

Btw, cassandra-medusa was installed using pip command `python3 -m pip install cassandra-medusa==0.22.3` 

Thanks in advance!!

### LINKED ISSUE #826: Backups for DSE 5.1 clusters fail since DSE support was added in 0.17.0
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=89340640)

Medusa 0.17.0 comes with the code changes introduced by #693 : _Support DSE (with search)_
The change, among other things, takes care of the `metadata` directory which was introduced in DSE 6.x.
However the code currently only differentiates between DSE and not DSE so it will try to perform a `metadata` snapshot even with DSE 5.1 and fail with:

```
[2024-11-29 00:58:43,054] INFO: Creating DSE snapshot
[2024-11-29 00:58:43,055] ERROR: Issue occurred inside handle_backup Name: fournodes01 Error: [Errno 2] No such file or directory: '/var/lib/cassandra/metadata'
[2024-11-29 00:58:43,055] DEBUG: Updated from existing status: 0 to new status: 2 for backup id: fournodes01
[2024-11-29 00:58:43,055] ERROR: Error occurred during backup: [Errno 2] No such file or directory: '/var/lib/cassandra/metadata'
```

the fix should be quite straightforward using `release_version` in addition to `server_type` when deciding if a 'DSE snapshot' is needed

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/MED-111) by [Unito](https://www.unito.io)
┆Issue Number: MED-111

### PR DESCRIPTION
**Pull request by arodrime**
## Summary

While testing on a real DSE 6.8 cluster, `backup-cluster` failed on every single node with:

```
[Errno 2] No such file or directory: '/myPathToData/dse/metadata'
```

`create_dse_snapshot()` assumes every DSE node has a `<dse_root>/metadata` folder to copy, but I *think* that's not always true (eg no DSE Management Services/Insights activity on that node). The `shutil.copytree` crash aborts `handle_backup` entirely - discarding all the SSTable upload work already done for that node and marking it failed.

Fix philosophy: treat a missing `metadata` folder as "nothing DSE-specific to snapshot" rather than a failure - same treatment on cleanup (`DseSnapshot.delete()`), which would otherwise crash on the same missing directory during `__exit__`. I hope that this is a honest response, not just the "easy" one. I have a (very) limited knowledge of DSE specifics.

On top of that, `_dse_root` was always *guessed* as "one directory up from the data folder", which breaks as soon as `metadata_directory` isn't at its default location (eg the data directory is moved to a dedicated disk). DSE actually writes this path itself into `cassandra.yaml` (key `metadata_directory`, see [DataStax docs](https://docs.datastax.com/en/dse/6.9/managing/configure/configure-cassandra-yaml.html)) - Medusa now reads it from there when present, and only falls back to the old guess for older DSE versions that never set the key (eg 5.1).

Fixes #826 - same crash, different trigger: #826 is DSE 5.1 never having a `metadata` folder at all, this case is a DSE 6.8 node that simply never populated one. Same crash site, same fix works for both regardless of the exact reason the folder is missing.

Fixes #819 - the metadata path is no longer guessed, it's read from `cassandra.yaml` the same way DSE itself defines it, so non-standard locations are now supported automatically instead of requiring a hardcoded guess.

## Tests
- `test_create_dse_snapshot_without_metadata_folder` - reproduces the exact crash without the fix, passes with it
- `test_create_dse_snapshot_with_metadata_folder` - existing behavior (folder present) unchanged
- `test_dse_metadata_directory_read_from_cassandra_yaml` - explicit `metadata_directory` in `cassandra.yaml` is used instead of the guess
- `test_dse_metadata_directory_falls_back_without_cassandra_yaml_key` - older DSE without the key keeps the previous guessed behavior

### DIFF
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -266,6 +266,13 @@ class CassandraConfigReader(object):
         return pathlib.Path(saved_caches_directory)
 
     @property
+    def dse_metadata_directory(self):
+        # DSE adds this key to cassandra.yaml itself (not dse.yaml). Older DSE versions (eg 5.1, #826)
+        # don't have it at all, so callers must fall back to a guessed location when this is None.
+        metadata_directory = self._config.get('metadata_directory')
+        return pathlib.Path(metadata_directory) if metadata_directory else None
+
+    @property
     def listen_address(self):
         if 'listen_address' in self._config and self._config['listen_address']:
             return self._config['listen_address']
@@ -355,8 +362,15 @@ class Cassandra(object):
         config_reader = CassandraConfigReader(cassandra_config.config_file, release_version)
         self._cassandra_config_file = cassandra_config.config_file
         self._root = config_reader.root
-        self._dse_root = self._root.parent
-        self._dse_metadata_folder = 'metadata'
+        dse_metadata_directory = config_reader.dse_metadata_directory
+        if dse_metadata_directory is not None:
+            self._dse_root = dse_metadata_directory.parent
+            self._dse_metadata_folder = dse_metadata_directory.name
+        else:
+            # metadata_directory isn't set in cassandra.yaml (older DSE, eg 5.1, #826) - guess its
+            # location next to the data directory, which holds for the common/default DSE layout.
+            self._dse_root = self._root.parent
+            self._dse_metadata_folder = 'metadata'
         self._commitlog_path = config_reader.commitlog_directory
         self._saved_caches_path = config_reader.saved_caches_directory
         self._hostname = contact_point if contact_point is not None else config_reader.listen_address
@@ -581,8 +595,8 @@ class Cassandra(object):
         return False
 
     def dse_snapshot_exists(self, tag):
-        # dse files live one directory up from the data folder
-        # the root field should point to the data directory as defined in the cassandra.yaml
+        # _dse_root is metadata_directory's parent (from cassandra.yaml) when set, otherwise a guess
+        # based on the data directory - see Cassandra.__init__.
         for snapshot in self._dse_root.glob(self.DSE_SNAPSHOT_PATTERN.format('*')):
             if snapshot.is_dir() and snapshot.name == tag:
                 return True
--- a/medusa/cassandra_utils.py
+++ b/medusa/cassandra_utils.py
@@ -463,9 +463,13 @@ class Cassandra(object):
 
         tag = "{}{}".format(self.SNAPSHOT_PREFIX, backup_name)
         if not self.dse_snapshot_exists(tag):
-            src_path = self._dse_root / self._dse_metadata_folder
-            dst_path = self._dse_root / self._dse_metadata_folder / 'snapshots' / tag
-            shutil.copytree(src_path, dst_path, ignore=Cassandra._ignore_snapshots)
+            src_path = self.dse_metadata_path
+            if src_path.is_dir():
+                dst_path = src_path / 'snapshots' / tag
+                shutil.copytree(src_path, dst_path, ignore=Cassandra._ignore_snapshots)
+            else:
+                # Not every DSE node has a metadata folder - warn instead of failing the backup.
+                logging.warning(f'No DSE metadata folder found at {src_path}, nothing to snapshot')
 
         return Cassandra.DseSnapshot(self, tag)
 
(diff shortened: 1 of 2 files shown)


rzvoncek (changes requested):
Since the metadata folder now can be anything, the hardcoded check [here](https://github.com/thelastpickle/cassandra-medusa/blob/radovan/configurable-chunks/medusa/cassandra_utils.py#L443) probably fails, meaning it does not do the recursion check properly. 

Could you please check if you could make it still prevent the cycle?

arodrime:
Good catch, thanks. 

`_ignore_snapshots` now takes the metadata folder name as a parameter instead of hardcoding 'metadata', so the recursion guard matches whatever metadata_directory actually resolves to. As unlikely as it is that someone would rename it, it was indeed a flow and risky/weak piece of code.

Added `test_create_dse_snapshot_with_custom_metadata_folder_name`, which reproduces the exact failure mode: a custom folder name, and asserts the snapshot doesn't recurse into its own snapshots dir. Also extended the existing `_ignore_snapshots` unit tests with a custom-name case and a mismatching-name case.

rzvoncek:
Thanks alain. Just one more thing. Could you please rebase on a master, again, and also run `poetry lock` to rebuild the lock file? I've merged a bunch of dependabot updates recently and this PR is not building now :(

arodrime:
Rebased on master. For poetry.lock, I went to regenerate it, but it turns out it wasn't needed, I think #979 already brought in the fixed lock file (see https://github.com/thelastpickle/cassandra-medusa/pull/979/changes#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89), so it just came along with the rebase. Ran the full test suite locally and 223 tests passed. 

Should be green now hopefully 🤞 .

Thanks for your time getting the stuff annoying my team into medusa mainstream, it will allow us to get back to the next released version.

rzvoncek:
Alain I'm sorrry but I had to cut a version out and this one didn't make it in. I can make you a patch version next week, hopefully.

arodrime:
> Alain I'm sorrry but I had to cut a version out and this one didn't make it in. I can make you a patch version next week, hopefully.

Ah! I didn't notice 😄 .

Fair enough; it took me time to get back to you. We can wait a week (or a few), no worries! I just wanted to say it is good that I won't have to keep a fork up to date in the long run, and thank you for the time you dedicated to our issues 🙏 .

PR #978, opened 2026-08-02
Increase multipart block size for azure from 4 to 20 MB

rzvoncek:


rzvoncek:
This was used for a one-off build, no intention of mergig. Superseeded by https://github.com/thelastpickle/cassandra-medusa/pull/979

PR #985, opened 2026-09-01
Replace parallel-ssh with asyncssh

### LINKED ISSUE #984: Replace parallel-ssh with asyncssh to remove gevent/greenlet dependency
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=237951883)

## Summary

Replace `parallel-ssh` with `asyncssh` as the SSH orchestration backend, removing `gevent`, `greenlet`, `ssh2-python`, and `ssh-python` from the dependency tree.

## Background

The current cluster orchestration ([medusa/orchestration.py](medusa/orchestration.py)) uses `parallel-ssh` (`pssh`) to run commands on Cassandra nodes in parallel. `parallel-ssh` has a hard dependency on `gevent`, which requires monkey-patching the entire Python standard library at process startup:

```python
# medusa/medusacli.py
from gevent import monkey
monkey.patch_all()
```

This patch must fire before any other imports because pssh uses gevent greenlets for all socket I/O. It replaces stdlib `socket`, `threading`, `time`, and others with gevent-aware equivalents. As a side effect, it also patches the `cassandra-driver`'s I/O layer, which has its own async reactor and does not require this treatment.

The five packages this pulls in (`parallel-ssh`, `gevent`, `greenlet`, `ssh2-python`, `ssh-python`) include compiled C extensions that require architecture-specific wheels, complicating packaging and container builds.

## Proposed change

Replace `parallel-ssh` with [`asyncssh`](https://asyncssh.readthedocs.io/), a pure-Python SSH2 implementation built on `asyncio` (stdlib). `asyncssh` supports all features currently used:

- Parallel execution across N hosts via `asyncio.gather()`
- PTY allocation
- SSH agent forwarding
- Certificate-based auth (`cert_file`)
- Private key auth
- Keepalive
- Per-host stdout/stderr/exit-code collection

## Migration scope

The change is contained to a single file: **`medusa/orchestration.py`** (~120 lines). The public interface — `pssh_run(hosts, command, hosts_variables, ssh_client)` — stays identical; callers in `backup_cluster.py` and `restore_cluster.py` do not change.

One small internal adjustment: pssh's `%s`/`host_args` per-host command substitution (used in the restore path in `restore_cluster.py`) would be replaced by pre-formatting the command string per host in Python before dispatch. This is a straightforward change since the substitution is already explained in a comment at the call site.

The `monkey.patch_all()` call in `medusacli.py` can be removed entirely once `parallel-ssh` is gone.

## Dependency delta

Removed (5 packages, including C extensions):
- `parallel-ssh`
- `gevent`
- `greenlet`
- `ssh2-python`
- `ssh-python`

Added (1 package, pure Python):
- `asyncssh` (depends only on `cryptography`, which is already a direct dependency)

## Files affected

- `medusa/orchestration.py` — rewrite SSH client using asyncssh
- `medusa/medusacli.py` — remove `gevent` import and `monkey.patch_all()`
- `pyproject.toml` — swap dependencies
- `tests/orchestration_test.py` — update mocks to match new client interface

rzvoncek:
* replaces parallel-ssh (and its deps like gevent and greenlet, which are just another way of doing async, on top of asyncio) with asyncssh, which is raw enough to use jusy asyncio
* adds an e2e tests that ssh-s between docker containers running actual sshd


rzvoncek:
Review.

1. Host key checking is now strict by default — blocking
  known_hosts empty (default) → orchestration.py:90-92 never passes known_hosts= to asyncssh.connect() → falls back to asyncssh's own default (~/.ssh/known_hosts, strict). Result: every host is refused, zero
  commands run.
  pssh_run returned: False
  ERROR: Host key is not trusted for host 127.0.0.1
  Not caught because orchestration_test.py:126-131 monkeypatches known_hosts=None, and the e2e fixture pre-populates known_hosts via ssh-keyscan. Need a documented way to disable checking (e.g. known_hosts = none
  → known_hosts=None), and it should be called out loudly in the changelog since it's default-on.

  2. use_sudo = True (the default) only prefixes the first command in the chain — blocking in practice
  orchestration.py:114-121 does 'sudo ' + cmd instead of wrapping the whole chain like pssh did (sudo -S $SHELL -c '<full cmd>'). For the real backup/restore command strings (mkdir -p {work}; cd {work} &&
  medusa-wrapper sudo medusa ...), only mkdir runs as root:
  sudo mkdir -p /tmp/medusa-job-abc; cd /tmp/medusa-job-abc && medusa-wrapper sudo medusa ...
  Work dir ends up root:root, then medusa-wrapper.sh (runs as the SSH user) fails to write stdout/stderr/pid/status there — permission error on every sudo'd backup/restore. Not caught because both e2e templates
  set use_sudo = False.

  Minor
  - No shlex.quote in the $SHELL -cl '...' wrapping — a single quote in a command breaks it (pssh had this too, worth fixing while touching the code).
  - forward_agent now defaults False; old code forced forward_ssh_agent=True — silent behavior change, worth a changelog line.
  - docs/Configuration.md doesn't document the new known_hosts/forward_agent options.
  - connect_timeout=30 hardcoded, not configurable.
  - packaging/docker-build/Dockerfile-{debian,debian-11,debian-12,ubuntu} still install gevent/greenlet — leftover, worth dropping since removing that dep chain was the point.
  - Stale "pssh" wording in medusacli.py:239, docs/Performing-backups.md, docs/Restoring-a-full-cluster.md.

PR #990, opened 2026-09-11
Ensure failing async operations with Azure close all FDs

### LINKED ISSUE #994: Azure storage leaks file descriptors
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=251413161)

burmanm:
We might leave some open in cases of errors and that leads to unnecessary open file descriptors.


tests/.DS_Store
rzvoncek: issue: could you please remove this file?

medusa/storage/azure_storage.py
```diff
@@ -192,7 +193,8 @@ async def _download_blob(self, src: str, dest: str):
             timeout=self.read_timeout,
         )
         Path(file_path).parent.mkdir(parents=True, exist_ok=True)
-        await downloader.readinto(open(file_path, "wb"))
+        with open(file_path, "wb") as f:
```
rzvoncek: To make Sonar happy, how about doing this like:

```python
        async with aiofiles.open(file_path, "wb") as f:
            async for chunk in downloader.chunks():
                await f.write(chunk)
```

rzvoncek (changes requested):

medusa/storage/azure_storage.py
burmanm: Yep.

rzvoncek (approved):

rzvoncek (approved):

PR #992, opened 2026-09-17
fix: BackupNode.tokens overflow

jaroug:
gRPC `BackupNode.tokens` overflow on clusters using `RandomPartitioner`: field changed from `int64` to `string` to support 128-bit tokens (breaking gRPC wire-format change for `medusa.proto`)

Note: this is totally vibe-coded


jaroug:
Fixes #993

